PR #2, opened 2017-06-10
Build with autotools+maven

rm5248:
This fix updates the log4cxx maven file to build with Maven properly when running on Linux.

It is rather simplistic at this point, and it will break on Windows.  There was some discussion earlier on using CMake for the build system instead of autotools.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,12 @@
 target/
+
+*.o
+*.lo
+Makefile
+Makefile.in
+*.m4
+.deps
+.dirstamp
+*.swp
+config**
+autom4te.cache
--- a/pom.xml
+++ b/pom.xml
@@ -173,19 +173,40 @@
 	<build>
 		<plugins>
 			<plugin>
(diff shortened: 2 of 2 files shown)


jgsogo:
Can you point me to the CMake discussion? I would be really interested on having a cross-platform solution.

tschoening:
There have been multiple discussions regarding CMake in the past, please search for them yourself on the mailing lists or in JIRA:

https://markmail.org/search/?q=list%3Aorg.apache.logging.dev+cmake
https://markmail.org/search/?q=list%3Aorg.apache.logging.log4cxx-dev+cmake

https://logging.apache.org/log4cxx/latest_stable/mail-lists.html

https://issues.apache.org/jira/browse/LOGCXX-486?jql=project%20%3D%20LOGCXX%20AND%20text%20~%20cmake

ams-tschoening:
It currently looks like #14 gets merged and using Autotools might get abandoned on Linux altogether, so I'm closing here for now.

PR #4, opened 2018-03-21
mbsrtowcs() returns the number of wide char,but not the source in str…

hustterry:
mbsrtowcs() returns the number of wide char,but not the source length.

If there have n chinese characters,the log will output more last n characters.

src/main/cpp/charsetdecoder.cpp
@@ -175,11 +175,11 @@ namespace log4cxx
                                stat = APR_BADARG;
                                in.position(src - in.data());
                                break;
                            } else {
                                stat = append(out, buf);
-                               in.position(in.position() + converted);
+                               in.position(in.position() + requested);
                            }
                       }
                   }
                   return stat;
               }

### DIFF
--- a/src/main/cpp/charsetdecoder.cpp
+++ b/src/main/cpp/charsetdecoder.cpp
@@ -177,7 +177,7 @@ namespace log4cxx
                                break;
                            } else {
                                stat = append(out, buf);
-                               in.position(in.position() + converted);
+                               in.position(in.position() + requested);
                            }
                       }
                   }


tschoening:
This has been reported before with different possible solutions. Is "requested" really correct compared to e.g. `sizeof(...) * converted` or such? Less than `requested` chars could have been processed in theory even without any error.

https://issues.apache.org/jira/browse/LOGCXX-369
https://issues.apache.org/jira/browse/LOGCXX-399

tschoening:
It's a string by definition so I guess we can ignore problems with terminating `0` within the string before `requested` has been reached.  Additionally, `x * converted` should be wrong as well because how many bytes one multibyte character spans depends on the current locale, right? So sounds safe to assume in case of success that simply all requested characters/bytes have been processed. This is reported to work in the linked issues as well:

> Then I modified the "in.position(in.position() + converted);" to "in.position(in.position() + requested);"

https://issues.apache.org/jira/browse/LOGCXX-399?focusedCommentId=13969295&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13969295

tschoening:
I fixed this using "requested" like mentioned in LOGCXX-399 and the linked PR #4 on GitHub. Additionally I added a test "decodingtest" using "encodingtest" as template, but the problem didn't occur using my C++Builder because I lack the used API, so am not sure if that works at all. It didn't fail before and doesn't now, so in the worst case is simply not doing anything useful. ;-)

PR #3, opened 2018-01-10
LOGCXX-494: windows cmake build for log4cxx against apr-2

jeking3:
https://issues.apache.org/jira/browse/LOGCXX-494

I've used log4cxx for quite some time and I was surprised to find there are no project or cmake files for windows builds in the trunk.  In the past I have cobbled together various sources and made visual studio projects to wrap apr, apr-util, and log4cxx into a single build.  

Today, I spent the day making a cmake build environment that leverages the latest apr-2 development trunk.

When you build apr-2 with cmake on Windows you get a static and a shared library in the prefix output directory.  The static library is named `apr-2.lib` and the dynamic link library is `libapr-2.{dll,lib}`.  I accounted for this in the cmake, expecting folks to set `-DBUILD_SHARED_LIBS={ON|OFF}`.  There are a number of build warnings when building dynamic, and it looks like some dllexport declarations are missing.  This PR does not correct those shortcomings.

I pulled some concepts from cmake files in the Apache Thrift project.

Here is an example configuration run and the output for configuring a Static, Release, VS2010 build.  The out-of-tree build directory is `c:\Temp\log4cxx-build\Static\Release`.  The source tree is `c:\temp\logging-log4cxx`.  Files will be installed into `C:\temp\log4cxx-install\Static\Release`.

In the build directory you get a full Visual Studio project that you can use if you want.

```
c:\Temp\log4cxx-build\Static\Release>cmake -G"Visual Studio 10 2010 Win64" c:\temp\logging-log4cxx -DCMAKE_INSTALL_PREFIX=C:\temp\log4cxx-install\Static\Release -DAPR2_ROOT_DIR=C:\Build\org.apache\apr\install\Static\Release
-- Configuring build environment for log4cxx version 0.11.1
-- The C compiler identification is MSVC 16.0.40219.1
-- The CXX compiler identification is MSVC 16.0.40219.1
-- Check for working C compiler: c:/Program Files (x86)/Microsoft Visual Studio 10.0/VC/bin/x86_amd64/cl.exe
-- Check for working C compiler: c:/Program Files (x86)/Microsoft Visual Studio 10.0/VC/bin/x86_amd64/cl.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: c:/Program Files (x86)/Microsoft Visual Studio 10.0/VC/bin/x86_amd64/cl.exe
[...]

PR #12, opened 2019-10-06
Add support for building and testing with cmake

swebb2066:
This just adds CMakeLists.txt files to various directories. No existing files are changed


ams-tschoening:
Thanks for the contribution, but I'm not sure how to deal with this currently and simply leave it open until further decisions. Some links to earlier discussions about that topic:

https://github.com/apache/logging-log4cxx/pull/2#issuecomment-370781162
https://github.com/apache/logging-log4cxx/pull/3#issuecomment-356979649

swebb2066:
To enable inclusion by vcpkg, log4cxx needs a modern cmake build system. I can see no problem having an one more way to build log4cxx. If there is some support I will add documentation etc.

jvz:
Can anything from this be combined with #14? I love the simplification if possible rather than supporting multiple build systems.

ams-tschoening:
From my understanding this is CMAKE without removing anything, while #14 is CMAKE with removing at least ANT, because in the past it has been discussed several times to focus on one build system only. So it's either about adding #12 or #14, not both I guess, with the latter making more sense in the long term anyway.

While we are at it, it should be discussed if CMAKE should replace Autotools on Linux as well, leaving only the ANT/MAVEN-parts in place necessary for generating the site, optionally release handling itself and stuff. Though especially the latter seems pretty complicated to me as well, but I'm not that familiar with MVN-releases.

jvz:
> While we are at it, it should be discussed if CMAKE should replace Autotools on Linux as well, leaving only the ANT/MAVEN-parts in place necessary for generating the site, optionally release handling itself and stuff. Though especially the latter seems pretty complicated to me as well, but I'm not that familiar with MVN-releases.

I like that idea.

rm5248:
Why use Maven at all at that point?  What I do for C++ projects is to just generate the site with Doxygen; it's possible to then make a doc/ subfolder that has the files in markdown or asciidoc([example](https://github.com/dbus-cxx/dbus-cxx/tree/7e48c593330a0b3bf9d576b8a35e4b74b944f602/doc)).  The style is different, but it would integrate nicely with the build process.

ams-tschoening:
> Why use Maven at all at that point? What I do for C++ projects is to just generate the site with Doxygen;[...]

Maven has been used for release-handling and site generation. Doxygen is used for apidocs as well already, but not for other parts of the site, like changes-reports and all that stuff.

ams-tschoening:
It currently looks like #14 gets merged, so I'm closing here for now.

PR #13, opened 2019-10-12
Add JSONLayout class

bblough:
This layout will format the LoggingEvent information such the date,
logger, level, message, and location information, as well as the values
in the MDC and NDC, into valid JSON.  This is useful for consumption
by elasticsearch or other log tools.

### DIFF
--- a/src/main/cpp/jsonlayout.cpp
+++ b/src/main/cpp/jsonlayout.cpp
@@ -1,0 +1,273 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <log4cxx/logstring.h>
+#include <log4cxx/jsonlayout.h>
+#include <log4cxx/spi/loggingevent.h>
+#include <log4cxx/level.h>
+#include <log4cxx/helpers/optionconverter.h>
+#include <log4cxx/helpers/iso8601dateformat.h>
+#include <log4cxx/helpers/stringhelper.h>
+
+#include <apr_time.h>
+#include <apr_strings.h>
+#include <string.h>
+
+using namespace log4cxx;
+using namespace log4cxx::helpers;
+using namespace log4cxx::spi;
+
+IMPLEMENT_LOG4CXX_OBJECT(JSONLayout)
+
+
+JSONLayout::JSONLayout() : locationInfo(false), dateFormat()
+{
+}
+
+
+void JSONLayout::setOption(const LogString& option, const LogString& value)
+{
+	if (StringHelper::equalsIgnoreCase(option,
+			LOG4CXX_STR("LOCATIONINFO"), LOG4CXX_STR("locationinfo")))
+	{
+		setLocationInfo(OptionConverter::toBoolean(value, false));
+	}
+}
+void JSONLayout::format(LogString& output,
+	const spi::LoggingEventPtr& event,
+	Pool& p) const
+{
+
+	output.append("{ ");
+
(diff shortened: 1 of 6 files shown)


src/main/cpp/jsonlayout.cpp
```diff
@@ -0,0 +1,273 @@
...
+	appendQuotedEscapedString(output, "message");
+	output.append(": ");
+	appendQuotedEscapedString(output, event->getMessage());
+
+	appendSerializedMDC(output, event);
+
+	appendSerializedNDC(output, event);
+
+
+	if (locationInfo)
```
ams-tschoening: How about putting the implementation into it's custom private `formatLocationInfo` or `fmtLocInfo` or alike? Would make `format` itself a bit shorter and easier to read. Appending the first `,` could stay here to make `fmtLocInfo` fully independent.

src/main/cpp/jsonlayout.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+	if (start < input.size())
+	{
+		buf.append(input, start, input.size() - start);
+	}
+
+	/* add trailing quote */
+	buf.push_back(0x22);
+
+	return;
```
ams-tschoening: You are mostly not using empty `return`s, so I suggest removing the current once.

src/main/cpp/jsonlayout.cpp
```diff
@@ -0,0 +1,273 @@
...
+			if (it + 1 != keys.end())
+			{
+				buf.append(", ");
+			}
+		}
+
+		buf.append(" }");
+	}
+
+	return;
```
ams-tschoening: Should be removed.

src/main/cpp/jsonlayout.cpp
```diff
@@ -0,0 +1,273 @@
...
+	logchar specialChars[] =
+	{
+		0x08,   /* \b backspace         */
+		0x09,   /* \t tab               */
+		0x0a,   /* \n newline           */
+		0x0c,   /* \f form feed         */
+		0x0d,   /* \r carriage return   */
+		0x22,   /* \" double quote      */
+		0x5c
+	}; /* \\ backslash         */
```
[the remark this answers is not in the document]
bblough: Hmm... That was originally on the line above.  I must've accidentally moved it when I realized I forgot the closing brace :-) Fixed.

src/main/cpp/jsonlayout.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+}
+
+void JSONLayout::appendSerializedMDC(LogString& buf,
+	const LoggingEventPtr& event) const
+{
+
+	LoggingEvent::KeySet keys = event->getMDCKeySet();
+
+	if (!keys.empty())
```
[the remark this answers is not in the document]
bblough: Fixed.

src/main/cpp/jsonlayout.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+}
+
+void JSONLayout::appendSerializedNDC(LogString& buf,
+	const LoggingEventPtr& event) const
+{
+
+	LogString ndcVal;
+
+	if (event->getNDC(ndcVal))
```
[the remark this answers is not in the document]
bblough: Fixed.

src/main/cpp/jsonlayout.cpp
bblough: Fixed.

src/main/cpp/jsonlayout.cpp
bblough: Fixed.

src/main/cpp/jsonlayout.cpp
bblough: I tried to stay consistent with what I already have and called it appendSerializedLocationInfo.  If you'd prefer me to rename everything to shorter or otherwise different names, let me know.  Fixed.

bblough:
For my use case, I think single-line log entries are preferred.  But I can see how other use cases might benefit from pretty-printed output.  I'll add it as a option.

bblough:
Pretty-print option added.  All changes pushed.

ams-tschoening:
Thanks for changing things, but I additionally ran into problems getting the new layout to compile in case the WCHAR_T-API is used and your tests failed on my Windows as well. The first problem could easily be changed by using `LOG4CXX_STR` everywhere and the latter by using `LOG4CXX_EOL` instead of `\n`. `HTMLLayout` seemed to have done the same.

Would be great if you could have a look at my newly created branch containing your PR and my additional fixes, if things still work for you.

https://github.com/apache/logging-log4cxx/tree/ghpr_13

bblough:
Ah, sorry about that.  Thanks for fixing it.  I've just compiled and tested here and it still works for me.

ams-tschoening:
That's good news, so I merged everything to master. Thanks for the contribution.

PR #14, opened 2020-01-31
Replace ant build with cmake

stephen-webb:


jvz:
I ❤️ PRs that have huge negative line counts like this! I'm not super familiar with the tooling in use, though I'll try this out over the weekend.

pom.xml
```diff
@@ -65,7 +65,7 @@
 		<connection>scm:git:http://gitbox.apache.org/repos/asf/logging-log4cxx.git</connection>
 		<developerConnection>scm:git:https://gitbox.apache.org/repos/asf/logging-log4cxx.git</developerConnection>
 		<url>https://gitbox.apache.org/repos/asf?p=logging-log4cxx.git</url>
-		<tag>HEAD</tag>
+		<tag>v0.11.0-RC1</tag>
```
jvz: This should be left as `HEAD` as the `mvn release:*` commands will handle setting the value for the appropriate tag commit.

jvz: Unless this PR has a staged release candidate it seems?

src/site/apt/building/cmake.apt
```diff
@@ -0,0 +1,75 @@
...
+
+  Make sure cmake 3.13+, g++ and make are available, install or
+  build apr 1.x, apr-util 1.x, gzip and zip.
+
++------------+
+$ apt-get install build-essential libapr1-dev libaprutil1-dev gzip zip
+$ cd apache-log4cxx-x.x.x
+$ mkdir build 
+$ cd build
+$ ccmake ..
```
jvz: Missing `cmake ..` step? I'm not very familiar with the tool, but `ccmake` seems to be the GUI to configure some options?

src/site/apt/building/cmake.apt
```diff
@@ -0,0 +1,75 @@
...
+| -DLOG4CXX_UNICHAR=yes  | Enable UniChar API methods, choice of yes, no (default).                                    |
+*------------------------+---------------------------------------------------------------------------------------------+
+| -DLOG4CXX_CFSTRING=yes | Enable CFString API methods, requires Mac OS/X CoreFoundation, choice of yes, no (default). |
+*------------------------+---------------------------------------------------------------------------------------------+
+
+* Platform specific notes:
+
+** Mac OS/X:
+
+  APR and APR-Util are provided by the platform in Mac OS/X 10.5 and iODBC in 10.4.
```
jvz: This might be obvious to you, but `cmake` does not come with macOS by default. It might help users to add the cmake dependencies explicit in the install commands given here for various operating systems.

CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
+cmake_minimum_required(VERSION 3.13)
```
rm5248: Is it possible to have a lower version of cmake?  What I generally do is to target the version of CMake that is in the oldstable version of Debian(as of right now, Debian 9(stretch) has cmake 3.7).  This is not a requirement.

CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+find_package(APR QUIET)
+# If APR find module sets the cache, the following will do nothing
+find_path(APR_INCLUDE_DIR apr.h PATH_SUFFIXES apr-1)
+find_library(APR_LIBRARIES NAMES libapr-1 apr-1)
+
+# Find Apache Runtime Utilities
+find_package(APR-Util QUIET)
+# If APR-UTIL find module sets the cache, the following will do nothing
+find_path(APR_UTIL_INCLUDE_DIR apu.h PATH_SUFFIXES apr-1)
+find_library(APR_UTIL_LIBRARIES NAMES libaprutil-1 aprutil-1)
```
rm5248: Shouldn't this use FindPkgConfig and pkg_check_modules on Unix systems instead of find_package?

CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+  FILES_MATCHING PATTERN "*.h"
+)
+
+install(TARGETS log4cxx EXPORT log4cxxTargets
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} 
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+)
```
rm5248: Similar to the above, on Linux at least GNUInstallDirs should be used instead, as that will take care of the multiarch specification on Debian-based systems.

CMakeLists.txt
stephen-webb: DO APR and APR-UTIL provide support pkg-config?

CMakeLists.txt
stephen-webb: I think I may have used some cmake that requires 3.13. It was 6 months ago so I cannot recall exact details.

I have been trying to follow best practices in Professional CMake (4th Edition)

CMakeLists.txt
rm5248: As far as I'm aware they do; I just tried this on Debian 10(buster), and it works properly though.  I'm not sure what is best.  Maybe this is better as a fallback if pkg_check_modules fails?

src/main/include/CMakeLists.txt
```diff
@@ -0,0 +1,101 @@
...
+  set(CHARSET_UTF8 1)
+endif()
+
+option(LOG4CXX_HAS_STD_LOCALE "Is the standard locale header available?" OFF)
+option(LOG4CXX_HAS_ODBC "Build with ODBC appender?" OFF)
+option(LOG4CXX_HAS_MBSRTOWCS "Default character encoder converts multi-byte string to LogString using mbstowcs()?" OFF)
+option(LOG4CXX_HAS_WCSTOMBS "Default wide character encoder converts using wcstombs()?" OFF)
+option(LOG4CXX_HAS_FWIDE "Is the fwide(fd) function available?" OFF)
+option(LOG4CXX_HAS_LIBESMTP "Use libESMTP in SMTPAppender?" OFF)
+option(LOG4CXX_HAS_SYSLOG "Is the syslog function available?" OFF)
```
rm5248: These should be checked automatically - see: https://github.com/rm5248/logging-log4cxx/blob/a3fd5a53aefb297dd4526acd156e11335e484c37/CMakeLists.txt#L94

rm5248:
Also: this doesn't install a pkgconfig file at all; that would be needed for other non-cmake buildsystems to work with.

pom.xml
stephen-webb: What is a staged release candidate?

CMakeLists.txt
stephen-webb: The best practices from Professional CMake book include the statement:
"Use of the FindPkgConfig module should be reserved only for those situations where find_package()
is not suitable."

I am inclined to stick with find_package if that is OK.

stephen-webb:
@rm5248 Can you tell me what I would need to change to install a pkgconfig file?

rm5248:
@stephen-webb You can configure the [.pc.in file](https://github.com/rm5248/logging-log4cxx/blob/a3fd5a53aefb297dd4526acd156e11335e484c37/CMakeLists.txt#L203).  [This is the pc.in file that I configured.](https://github.com/rm5248/logging-log4cxx/blob/a3fd5a53aefb297dd4526acd156e11335e484c37/liblog4cxx.pc.in)

src/site/apt/building/cmake.apt
stephen-webb: Could you tell what to add as I do not have macOS

src/site/apt/building/cmake.apt
stephen-webb: Do you think I should duplicate the Windows instructions for building from source into a unix/macOS section (just need to change \ to / actually)?

src/site/apt/building/cmake.apt
jvz: The most common installer is via `brew install cmake` using [homebrew](https://brew.sh/) or it can always be downloaded from [their binaries page](https://cmake.org/download/).

src/site/apt/building/cmake.apt
jvz: If there's any differences in the commands, go ahead.

stephen-webb: Unfortunately, the Windows technique will not work on unix/macOS as APR and APR-Util CMakeLists.txt file are Windows-only implementations.

stephen-webb:
Hi All,

I have done everything I can think of prepare
https://github.com/apache/logging-log4cxx/pull/14 for inclusion in the
repository.

Please review.

ams-tschoening:
So let's close #12 and really focus on this one, replacing things in favour of CMAKE instead of adding it? I don't use the ANT-build myself anyway, so would focus on that site generation etc. still works. It's just that things might break for other users.

src/test/cpp/net/socketserverstarter.cpp
```diff
@@ -0,0 +1,108 @@
...
+          LOGUNIT_ASSERT(stat == APR_SUCCESS);
+          // convert the space separated cmd string to the argument list
+          //
+          char** args = (char**)apr_palloc(pool, 15 * sizeof(*args));
+          char* pcmd = apr_pstrdup(pool, cmd);
+          int i = 0;
+          for (; i < 14 && pcmd && *pcmd; ++i)
+          {
+              args[i] = pcmd;
+              if (NULL != (pcmd = strchr(pcmd, ' ')))
```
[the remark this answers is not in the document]
ams-tschoening: `;` is difficult on Windows, because it's used as divider for the Java-classpath already, while `:` is difficult on non-Windows, because that is used as divider for Java-classpaths. Might be best to not embed the classpath into the command at all and use [CLASSPATH](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/classpath.html) environment variable instead?

src/test/cpp/net/socketservertestcase.cpp
```diff
@@ -478,4 +477,4 @@ LOGUNIT_CLASS(SocketServerTestCase)
 const File SocketServerTestCase::TEMP("output/temp");
 const File SocketServerTestCase::FILTERED("output/filtered");
 
-LOGUNIT_TEST_SUITE_REGISTRATION_DISABLED(SocketServerTestCase)
+LOGUNIT_TEST_SUITE_REGISTRATION(SocketServerTestCase)
```
ams-tschoening: Should this test really be enabled? While your CMAKE-build might handle things automatically, it might be more difficult for users without CMAKE. Took me a while to figure things out based on this commit myself and the test still fails currently...

	set PATH=C:\Program Files\Java\jdk-8\bin;C:\Program Files (x86)\UnxUtils\usr\local\wbin;%PATH%
	set TEST_PROJ=%cd%
	set LOG4CXX_VER_BASE=..\..\..\..
	set TEST_SRC=%LOG4CXX_VER_BASE%\src\src\test
	set TEST_RES=%TEST_SRC%\resources

	set TOTO=wonderful
	set key1=value
	set key2=value2

	set log4j_CLASSPATH=%LOG4CXX_VER_BASE%\..\..\..\..\Java\log4j-1.2.17.jar
	set SOCKET_SERVER_SOURCES=%TEST_SRC%\java\org\apache\log4j\net\ShortSocketServer.java
	set SOCKET_SERVER_CLASSPATH=%TEST_RES%;%log4j_CLASSPATH%
	set CLASSPATH=%SOCKET_SERVER_CLASSPATH%
	set SOCKET_SERVER_COMMAND=C:\Program Files\Java\jdk-8\bin\java.exe;org.apache.log4j.net.ShortSocketServer;8;input/socketServer

	pushd "%TEST_RES%"
	javac -d "." -classpath "%log4j_CLASSPATH%" "%SOCKET_SERVER_SOURCES%"
	"%TEST_PROJ%\Win32\Debug\out\%~n0.exe" %*
	popd

	goto :EOF

ams-tschoening:
@stephen-webb Does `socketservertestcase` succeed for your? I get the following error messages for two tests which use the `%C` format flag in their configs:

    |Files [output/filtered] and [witness/socketServer.2] differ on line 1
    One reads:  [TRACE T2 [main] ? (socketservertestcase.cpp:XXX) Message 1].
    Other reads:[TRACE T2 [main] SocketServerTestCase (socketservertestcase.cpp:XXX) Message 1].
    --------------------------------
    Contents of output/filtered:
    1   : TRACE T2 [main] ? (socketservertestcase.cpp:XXX) Message 1
    [...]
    --------------------------------
    Contents of witness/socketServer.2:
    1   : TRACE T2 [main] SocketServerTestCase (socketservertestcase.cpp:XXX) Message 1

I have no idea where that `?` comes from...

src/site/apt/building/cmake.apt
```diff
@@ -0,0 +1,104 @@
...
+$ cd apache-log4cxx-x.x.x
+$ mkdir build 
+$ cd build
+$ ccmake ..
+$ make
+$ sudo make install
++------------+
+
+* ccmake options
+*------------------------+---------------------------------------------------------------------------------------------+
```
ams-tschoening: A newline is necessary BEFORE the table or it doesn't render properly and is instead interpreted as additional text for the headline `ccmake options` .

Clipboard02

src/site/apt/building/vcpkg.apt
```diff
@@ -0,0 +1,41 @@
...
+   Then, to hook up user-wide integration, run (note: requires admin on first use)
+
+   PS> .\vcpkg integrate install
+   Linux:~/$ ./vcpkg integrate install
++----+
+
+*Building log4cxx.dll
+
+   PS> .\vcpkg install log4cxx
+   Linux:~/$ ./vcpkg install log4cxx
```
ams-tschoening: These two lines don't seem to render as expected:

Clipboard02

I suggest simply using `+----+` like above?

ams-tschoening:
https://github.com/apache/logging-log4cxx/tree/ghpr_14_replace-ant-build-with-cmake

src/test/cpp/net/socketservertestcase.cpp
stephen-webb: Disabling a testcase is needed when building multiple tests into a single executable. To run a disabled test case you then need to add it as a parameter.
The socketservertestcase excutable has only one test case so I do not think it is appropriate to disable it.

src/test/cpp/net/socketserverstarter.cpp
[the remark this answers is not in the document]
stephen-webb: To avoid issues with ';' processing, I have added the capability to load the command parameters from a file (one per line) into socketserverstarter. 
I now use this technique when running the socketservertestcase from ctest.
I have left the SOCKET_SERVER_COMMAND as space separated with simple support for quoted arguments

src/site/apt/building/cmake.apt
stephen-webb: The socketservertestcase fails on Windows for me too. I have not investigated the cause.
It also (sometimes) fails on Linux (Ubuntu) when run from Maven. It always seem to succeed on Linux when I run ctest from the command prompt.

src/test/java/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+)
+FetchContent_GetProperties(log4j)
+if(NOT log4j_POPULATED)
+  FetchContent_Populate(log4j)
+endif()
+set(log4j_CLASSPATH "${log4j_SOURCE_DIR}/log4j-1.2.17.jar" )
+set(SOCKET_SERVER_SOURCES org/apache/log4j/net/ShortSocketServer.java)
+add_custom_target(test-classes
+  COMMAND ${Java_JAVAC_EXECUTABLE} -d ${CMAKE_CURRENT_BINARY_DIR}
+          --class-path "${log4j_CLASSPATH}" ${SOCKET_SERVER_SOURCES}
```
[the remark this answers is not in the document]
stephen-webb: I have changed it to -classpath

src/test/cpp/net/socketserverstarter.cpp
```diff
@@ -0,0 +1,108 @@
...
+#ifdef SHELL_CMD_TYPE_WORKS
+          stat = apr_procattr_cmdtype_set(attr, APR_SHELLCMD);
+          LOGUNIT_ASSERT(stat == APR_SUCCESS);
+          stat = apr_proc_create(&server_pid, cmd, NULL, NULL, attr, pool);
+#else
+          stat = apr_procattr_cmdtype_set(attr, APR_PROGRAM);
+          LOGUNIT_ASSERT(stat == APR_SUCCESS);
+          // convert the space separated cmd string to the argument list
+          //
+          char** args = (char**)apr_palloc(pool, 15 * sizeof(*args));
```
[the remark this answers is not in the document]
stephen-webb: I nave replaced the number with a named constant

src/site/apt/building/cmake.apt
stephen-webb: Six test cases (including socketservertestcase) require the 'sed' program to pass.

I have now excluded those when the 'sed' program is not available.

src/test/cpp/net/socketservertestcase.cpp
ams-tschoening: The point is not about different executables, but that this one test case only requires additional Java and is difficult to setup. It has been disabled in the past for a long time already:

https://github.com/apache/logging-log4cxx/commit/b279aa7922fe8cbdc3ab2da524e1d6a9ed3913bb#diff-78b9ec5d590a88f742877f2f096c52a9R51

src/test/cpp/net/socketserverstarter.cpp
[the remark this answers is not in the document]
ams-tschoening: `SOCKET_SERVER_COMMAND` now works for me using the following line:

    set SOCKET_SERVER_COMMAND="C:\Program Files\Java\jdk-8\bin\java.exe" -classpath "%SOCKET_SERVER_CLASSPATH%" org.apache.log4j.net.ShortSocketServer 8 input/socketServer

ams-tschoening:
> I have no idea where that `?` comes from...

I think the log settings and witnesses were simply wrong, because the test has been disabled so long. `%C` gets the issuing class of a log statement, which might not be possible at all if a message is sent over network. `%c` works perfectly instead, maybe `%C` had a different fallback than `?` in the past? the following commit describes the changes after which all tests succeed for me:

https://github.com/apache/logging-log4cxx/commit/0baf6074cbfcf40026b4e41d25734944cb58df7c

I've additionally applied reformatting to the code, because non-tests have been reformatted in the past already and the tests were pretty inconsistent with indentation and stuff. Things look good to merge for me now. I won't have the time to test the CMAKE-stuff myself, but problems can be solved in master in future as well.

src/test/cpp/net/socketserverstarter.cpp
```diff
@@ -0,0 +1,108 @@
...
+   }
+   
+   void startServer()
+   {
+     helpers::Pool p;
+     apr_pool_t* pool = p.getAPRPool();
+     char* cmd = NULL;
+     apr_status_t stat = apr_env_get(&cmd, "SOCKET_SERVER_COMMAND", pool);
+
+     if (cmd && *cmd)
```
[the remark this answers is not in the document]
ams-tschoening: Changed things myself: https://github.com/apache/logging-log4cxx/commit/e8fe4890beee121adbd1a78969f792af4b0ad747

ams-tschoening:
Using `com.googlecode.cmake-maven-project` with OpenJDK 8 fails, because they [require JDK 11](https://github.com/cmake-maven-project/cmake-maven-project/commit/715e2c8775a3bc63dbca55911e5701444d0c0970) for around a year now. [Version 3.7.2](https://github.com/cmake-maven-project/cmake-maven-project/blob/release-3.7.2-b1/.travis.yml) seems to be the last building with JDK 8.

Targetting older versions of CMAKE has been [asked before](https://github.com/apache/logging-log4cxx/pull/14/files/af0134afaf2cd8d85d60aca3bae4b627d57b8c48#r373735482) as well and 3.7 was the preferred target then as well.

So, is it really necessary to use these current versions instead of older ones? Are the Maven plugin and the version of CMAKE bound at each other? Simply downgrading the MVN-plugin doesn't work:

    [ERROR] Failed to execute goal com.googlecode.cmake-maven-project:cmake-maven-plugin:3.7.2-b1:generate (cmake-generate) on project apache-log4cxx: The parameters 'generator' for goal com.googlecode.cmake-maven-project:cmake-maven-plugin:3.7.2-b1:generate are missing or invalid -> [Help 1]

https://github.com/cmake-maven-project/cmake-maven-project/blame/release-3.7.2-b1/README.md#L30

stephen-webb:
The previous version of cmake.maven.plugin did not work on Windows.
The maintainer is active so may be receptive if you raise an issue.

ams-tschoening:
I'm trying to test CMAKE and MVN on Sunday myself, just out of interest to see what happens, but nevertheless, as long as nobody objects, going to merge `ghpr_14_replace-ant-build-with-cmake` anyway. That contains this PR entirely, but some additional code style fixes as well.

So until Sunday, people should base their work on `ghpr_14_replace-ant-build-with-cmake`, like #16 was, and afterwards things simply need to be fixed as necessary in `master`.

PR #18, opened 2020-02-17
Conditionally exclude tests that require the external 'zip' or 'gzip'

stephen-webb:
Exclude tests that require  'zip' or 'gzip' when the program 'zip' or 'gzip' is not available

### DIFF
--- a/src/cmake/FindAPR-Util.cmake
+++ b/src/cmake/FindAPR-Util.cmake
@@ -38,7 +38,7 @@ if(EXISTS ${APR_UTIL_CONFIG_EXECUTABLE})
 else()
     find_path(APR_UTIL_INCLUDE_DIR apu.h PATH_SUFFIXES apr-1)
     find_library(APR_UTIL_LIBRARIES NAMES libaprutil-1 aprutil-1)
-    find_path(APR_UTIL_DLL_DIR libaprutil-1.dll)
+    find_program(APR_UTIL_DLL libaprutil-1.dll)
 endif()
 
 find_package_handle_standard_args(apr-util DEFAULT_MSG
--- a/src/cmake/FindAPR.cmake
+++ b/src/cmake/FindAPR.cmake
@@ -37,7 +37,7 @@ if(EXISTS ${APR_CONFIG_EXECUTABLE})
 else()
     find_path(APR_INCLUDE_DIR apr.h PATH_SUFFIXES apr-1)
     find_library(APR_LIBRARIES NAMES libapr-1 apr-1)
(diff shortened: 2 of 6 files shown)


stephen-webb:
I have not been through all the tests that use sed in as much detail as
timebasedrollingtestcase. It takes quite a lot of time to understand what
each test is doing.
The two rolling test cases failed on Windows as gzip and zip are not
available. I spent today getting to an understanding of the problem. Still
not fully aware of why the non-empty output directory caused failures.

I will not be able to do more until next weeekend now.

ams-tschoening:
So what's the problem actually, are you not able to provide the necessary binaries for some reason or do you want to make tests more compatible with systems not having them by default?

Like already said, SED, ZIP and GZIP can easily be provided on Windows, so maybe the better approach is to document that those are needed if not already the case. You didn't care about Java not being available for [socketservertestcase](https://github.com/apache/logging-log4cxx/pull/14/files/d0c6cae2e4ede3cc6530f4b5ac8f80732a6b08fc..b971dd0bba2c548539aefb60ca6b1ff0e2bbf10e#diff-4360b6f15ada2e89dcba704aecb337f4) as well. These binaries are required by many tests and excluding each of those doesn't feel right.

I don't think we should do this and vote against merging this. Additionally, the SED-related exception should be removed as well if this doesn't get merged.

stephen-webb:
Tests fail on Windows at the moment. I believe that is not good enough for
release. People will not be confident using the release.
I am working to get all tests passing.

ams-tschoening:
I'm running all tests successfully already on Windows, so if that is the main problem, please create an issue with all the failures you run into and we solve them. I can tell you about which binaries are additionally required. That can afterwards be documented instead of excluding tests.

> C:\Program Files (x86)\UnxUtils\usr\local\wbin
https://sourceforge.net/projects/unxutils/

The problem with excluded tests additionally is that people think that things are working as expected, if in reality many tests might simply not have been run at all because of missing requirements.

ams-tschoening:
I've removed the SED-exceptions to be in-line with my former argument and am closing here for now as well. SED etc. are used with many different tests and providing those under Windows is easy, so we need better arguments to exclude tests than only that they are not available.

The one exception I'm currently unsure about is `Java_Development_FOUND`, because Java is a completely different ecosystem and stuff. But the goal should be to be able to run most other tests until bars really get too high, which I don't see with SED and Co. Instead that there are necessary and where to get them should be documented somewhere or one can simply tell people.

@stephen-webb, please create some distinct issue with the APR-problems you ran into so that we can discuss those independently.

PR #20, opened 2020-03-09
Windows build static library support

stephen-webb:


ams-tschoening:
If I reviewed correctly, there are multiple issues with this PR:

1. it seems to reintroduce excluding tests, which I already removed in the past. Providing the missing binaries on Windows is easy and as many tests should be executed as possible in general. I won't merge excluding tests without any further discussion on why exactly you want those tests to be excluded. The binaries not being available by default in itself is not enough.

2. This PR seems to contain #19 as well, for which I asked for additional changes on its own already. So please reduce this PR to what is really necessary for static library support itself only. Thanks!

stephen-webb:
The PR may have picked up some other commits. I will look at it tomorrow.

ams-tschoening:
Closing in favour of #21.

PR #21, opened 2020-03-10
Add support for building log4cxx as a statically linked library on Windows

stephen-webb:
Replacement for PR #20


ams-tschoening:
Any reason why `-DAPU_STATIC` and `-DAPR_STATIC` are not documented with something like `=off|on` , like `-DBUILD_SHARED_LIBS` is? Does this work differently or ...?

stephen-webb:
No, they work the same, except the Build shared is on by default. The =on
does no more than the existence of a definition.

stephen-webb:
Correction, =yes is equivalent to existence of a definition

ams-tschoening:
Then please add `=yes|no|on|off` as necessary/useful for `-D*_STATIC`, so that its in line with the other documented options. Afterwards I'm going to merge.

PR #22, opened 2020-04-09
Only exporting the log4cxx symbols by default

aallrd:
Adding a linker version script to only make the log4cxx symbols visible in the log4cxx shared library.
The exported defined symbols on the installed library can be checked with the below command:
```
nm -gDC --defined-only /usr/local/lib64/liblog4cxx.so
```
Number of exported symbols other than logc4xx ones **before** the PR:
```
nm -gDC --defined-only /usr/local/lib64/liblog4cxx.so | grep -v log4cxx | wc -l
336
```
Number of exported symbols other than logc4xx ones **after** the PR:
```
nm -gDC --defined-only /usr/local/lib64/liblog4cxx.so | grep -v log4cxx | wc -l
0
```


rm5248:
I'm not all that familiar with this, so out of curiosity what problem does this attempt to solve?  I've never had an issue with this on any library before, but is there some use case that I'm not aware of?

It would also probably make sense for this to be off by default, as it looks like the linker script will only work with g++/clang++?  That could prove problematic on Windows.

ams-tschoening:
I was about to ask the same question about what problem the PR actually solves and which compilers it supports.

CMakeLists.txt
```diff
@@ -18,6 +18,13 @@ if(NOT BUILD_SHARED_LIBS)
   set(LOG4CXX_COMPILE_DEFINITIONS LOG4CXX_STATIC)
 endif()
 
+option(EXPORT_ONLY_LOG4CXX_SYMBOLS "Only exports the log4cxx symbols" ON)
+if(BUILD_SHARED_LIBS AND EXPORT_ONLY_LOG4CXX_SYMBOLS)
+  SET(LINKER_VERSION_SCRIPT_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/main/resources/log4cxx.map")
+  SET(LINKER_VERSION_SCRIPT_FLAG "-Wl,--version-script=${LINKER_VERSION_SCRIPT_FILE}")
```
ams-tschoening, line 24: Does this work for any compiler supported by CMAKE? The arguments look pretty compiler-specific to me.

CMakeLists.txt
```diff
@@ -18,6 +18,13 @@ if(NOT BUILD_SHARED_LIBS)
   set(LOG4CXX_COMPILE_DEFINITIONS LOG4CXX_STATIC)
 endif()
 
+option(EXPORT_ONLY_LOG4CXX_SYMBOLS "Only exports the log4cxx symbols" ON)
```
ams-tschoening, line 21: What's the benefit of enabling this by default? Especially if it's unclear currently if different compilers are supported.

src/site/apt/building/cmake.apt
```diff
@@ -39,24 +39,26 @@ $ sudo make install
...
+|                                   | Any compilation unit that includes a log4cxx header must define LOG4CXX_STATIC.             |
+*-----------------------------------+---------------------------------------------------------------------------------------------+
+| -DAPU_STATIC=yes                  | Link to the APR-Util static library. By default, the log4cxx shared library is linked to the|
+|                                   | APR-Util shared library. If BUILD_SHARED_LIBS=off, the static APR-Util library is used.     |
+*-----------------------------------+---------------------------------------------------------------------------------------------+
+| -DAPR_STATIC=yes                  | Link to the APR static library. By default, the log4cxx shared library is linked to the     |
+|                                   | APR shared library. If BUILD_SHARED_LIBS=off, the static APR library is always used.        |
+*-----------------------------------+---------------------------------------------------------------------------------------------+
+| -DEXPORT_ONLY_LOG4CXX_SYMBOLS=yes | Only export the log4cxx symbols in the log4cxx shared library.                              |
```
ams-tschoening, line 60: Maybe I'm wrong, but the other docs seem to document the non-default values. So this should be `no` instead? Things should made be clear in the text description as well:

> -DEXPORT_ONLY_LOG4CXX_SYMBOLS=no
> Only export the log4cxx symbols in the log4cxx shared library, choice of yes (default), no.

aallrd:
Hello,

Sorry for the incomplete PR, I wanted really to open an issue but I could not find how to to that on this project.

The goal of this change is to limit the visibility of the symbols of the log4cxx shared library (https://gcc.gnu.org/wiki/Visibility).
The trick here is that instead of modifying the code to add the proper visibility attributes we let the linker do the job, this is less optimal but easier.

I am actually using this change in my project to "fix" a performance issue I observed recently.
I am using the log4cxx shared library in release mode (O3) in my project.
A build mistake put the log4cxx shared library as a dependency of another shared library that did not need it, and even though it was not using it at all there was a performance gain observed in the tests.
Looking at a profiler report, I noticed that some standard functions and templates (like strings or others) were actually relocated at runtime to the log4cxx shared library (using `LD_DEBUG=bindings LD_DEBUG_OUTPUT=ld.log ldd -r app`).

I was not expecting this relocation to happen, so I started reading about symbols visibility and exports in shared libraries. That lead me to this GNU linker option `--version-script` to select which symbols to export globally/locally.
Using this option on my internal build of log4cxx allowed me to go back to my previous "slow" behavior and non-log4cxx symbols not being looked up in the log4cxx shared library.

I am not sure if this behavior is expected or not, I could not find a clear answer on that topic yet.

I think the same behavior can be achieved on Windows with Visual Studio using the linker [/DEF option](https://docs.microsoft.com/en-us/cpp/build/reference/module-definition-dot-def-files?view=vs-2019).

CMakeLists.txt
aallrd: This is a linker option, supported by GNU ld and GNU gold.

CMakeLists.txt
aallrd: I put the default to false.

src/site/apt/building/cmake.apt
aallrd: Updated the documentation to show the default (no).

rm5248:
In that case, there's probably an easier way to do this.  Since the symbols already need to be marked as exported for Windows builds, they already have a `LOG4CXX_EXPORT` macro associated with them.  This is not set to anything on [Linux/OSX builds](https://github.com/apache/logging-log4cxx/blob/671f0b5663b6b1063cfaa44acaee1b244fec3769/src/main/include/log4cxx/log4cxx.h.in#L46), but since the macro is there all you _should_ have to do is modify log4cxx.hw.in to look something like the following:

```
#define LOG4CXX_EXPORT_SYMBOLS_ONLY @EXPORT_ONLY_LOG4CXX_SYMBOLS@

#if LOG4CXX_EXPORT_SYMBOLS_ONLY
  #if __GNUC__ >= 4
      #define LOG4CXX_EXPORT __attribute__((visibility("default")))
  #else 
      #define LOG4CXX_EXPORT
  #endif
#else
  #define LOG4CXX_EXPORT
#endif
```

And then add to the compilation flags `-fvisibility=internal`, something like this:

```
if(BUILD_SHARED_LIBS AND EXPORT_ONLY_LOG4CXX_SYMBOLS)
  check_cxx_compiler_flag("-fvisibility=internal" LOG4CXX_SUPPORTS_VISIBILITY_INTERNAL)
  if( LOG4CXX_SUPPORTS_VISIBILITY_INTERNAL )
    set( LOG4CXX_COMPILE_OPTIONS ${LOG4CXX_COMPILE_OPTIONS} "-fvisibility=internal")
  endif( LOG4CXX_SUPPORTS_VISIBILITY_INTERNAL )
endif()

# ... in src/CMakeLists.txt
target_link_libraries(log4cxx PRIVATE ${LOG4CXX_COMPILE_OPTIONS})
```

Weather or not that will work appropriately/is a good idea I don't know.

ams-tschoening:
@aallrd

> Sorry for the incomplete PR, I wanted really to open an issue but I could not find how to to that on this project.

Issues are maintained in the [Apache JIRA](https://issues.apache.org/jira/projects/LOGCXX/issues/) by default. so either create one there first next time or simply be a bit more verbose in the initial PR already. Doesn't make too much difference in my opinion.

> Looking at a profiler report, I noticed that some standard functions and templates (like strings or others) were actually relocated at runtime to the log4cxx shared library (using LD_DEBUG=bindings LD_DEBUG_OUTPUT=ld.log ldd -r app).

Do you know which functions and templates and why exactly that happens? I wonder if that is expected at all already and should be fixed instead of working around it using the symbols?

> I think the same behavior can be achieved on Windows with Visual Studio using the linker /DEF option.

Just as some docs, because I'm using C++-Builder, the following should be the corresponding options for that?

* http://docwiki.embarcadero.com/RADStudio/Rio/en/Module_Definition_Files
* http://docwiki.embarcadero.com/RADStudio/Rio/en/IMPDEF.EXE,_the_Module_Definition_Manager
* http://docwiki.embarcadero.com/RADStudio/Rio/en/MKEXP.EXE,_the_64-bit_Windows_Import_Library_Tool_for_C%2B%2B
* http://docwiki.embarcadero.com/RADStudio/Rio/en/C%2B%2B_Linker_Output

@rm5248 

> In that case, there's probably an easier way to do this.[...]

I'm just trying to understand things, so: Why do you consider your approach easier? Having a well documented cmake-option doesn't seem that wrong to me. Your approach only implements non-Windows-compilers, doesn't it? Might the problem arise that this PR needs to be modified for Windows-compilers as well at some point? In the latter case it might be easier to add that to CMAKE only with different compiler-options?

CMakeLists.txt
ams-tschoening: So `log4cxx.map` is linker-specific and might be named more specifically? Something like `log4cxx.gnu[_ld].map` or something like that? You already mentioned visual studio, so something like `log4cxx.vs.def` might need to be added in future as well? Do all those files share some common name one could reuse, something about symbols or exporting or linking or ...?

ams-tschoening:
@aallrd How about updating your branch to master and including the attached changes to `changes.xml` in this PR? I've changed that file yesterday to be able to link to PRs automatically.

[ghpr_22.patch.txt](https://github.com/apache/logging-log4cxx/files/4466165/ghpr_22.patch.txt)

rm5248:
> I'm just trying to understand things, so: Why do you consider your approach easier? Having a well documented cmake-option doesn't seem that wrong to me. Your approach only implements non-Windows-compilers, doesn't it? Might the problem arise that this PR needs to be modified for Windows-compilers as well at some point? In the latter case it might be easier to add that to CMAKE only with different compiler-options and/or `.map|.def`-files?

My understanding(which may be incorrect) is that on Windows, the `__declspec(dllexport)` does the same thing as `__attribute__((visibility("default")))` with g++, putting the symbol into the symbol table of the object file. The main difference seems to be that g++ and/or ELF files by default export all symbols, e.g. `__attribute__((visibility("default")))` is set automatically.

Using the `-fvisibility=internal` does not export the symbols by default(similar to Windows).  So if we turn this on in the compiler, we just need to determine which symbols are exported, which should be accomplished by setting the `LOG4CXX_EXPORT` macro.

It's simpler in my mind since that means there's no linker script that needs to be modified, although I believe that the end result will be the same.

I don't _think_ that anything needs to be done on Windows(due to my understanding of how `__declspec(dllexport)` works), but I don't have enough knowledge of how Windows works to say for certain. 

[Windows documentation](https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=vs-2019)

---

After doing some tests on my system however, it seems that just setting `-fvisibility=hidden` does not seem to remove the other libstdc++ symbols(see branch [here](https://github.com/rm5248/logging-log4cxx/commit/f36302e7c7c82efced0f8fd721416a14c732bf5b)).

There's also this bug on the [GCC bugzilla](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36022) that addresses the symbol exporting.

Ultimately I don't really know what the correct solution is here, but I suspect that only telling the linker file to export nothing but `log4cxx` symbols may cause undefined behavior and/or linker issues in the future.

ams-tschoening:
So what's the strategy then? @aallrd Do you want to give the approach of @rm5248 a try and see if it fixes your concrete problem as well? Having only one concept to deal with the same thing would be of benefit of course. IF it's the same concept and can handle both things...

OTOH, the changed implementation of CMAKE doesn't look too invasive and can easily be disabled if it makes problems in future by all users. So there doesn't seem to be too much risk merging it?

rm5248:
> OTOH, the changed implementation of CMAKE doesn't look too invasive and can easily be disabled if it makes problems in future by all users. So there doesn't seem to be too much risk merging it?

Since it is off by default, I don't think that it will cause an issue for normal users, so I agree that the risk is low.

aallrd:
Hello,
Thank you for your comments, and for the interesting [GCC bugzilla report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36022).
@rm5248, you are right that setting the visibility using these macros should be the best thing to do, as documented in the [GCC Visibility](https://gcc.gnu.org/wiki/Visibility) page.
It makes the code itself cross-platforms and enables better compiler optimization.
Regarding the linker version script, after reading even more on the subject I found that it can be indeed dangerous in some situations and that some std symbols are exported on purpose in order to allow throw-catch/dynamic_cast to work across shared libs boundaries.
In the end I am not sure what to do anymore, and I am not convinced that this change is a good one. Since I don't fully understand either what is going on, I would like to avoid breaking things for you :).
I will keep this patch on my internal project for the moment, try to implement the visibility macros for GCC in the project and test if they help solve my initial issue as well.

ams-tschoening:
> I will keep this patch on my internal project for the moment, try to implement the visibility macros for GCC in the project and test if they help solve my initial issue as well.

Sounds like a good idea. Than please report back if you have any additional findings and until then I simply won't merge, but keep this open for some time. Feel free to close it yourself if you think that's necessary.

rm5248:
> I will keep this patch on my internal project for the moment, try to implement the visibility macros for GCC in the project and test if they help solve my initial issue as well.

@aallrd  If you didn't see it before, I've got a patch [here](https://github.com/rm5248/logging-log4cxx/commit/f36302e7c7c82efced0f8fd721416a14c732bf5b) that should implement the macros appropriately.  It should be a good starting point for you.

ams-tschoening:
A year has passed without further activity, so time to close for now. In case of new conclusions, feel free to reopen to keep the history of this concrete discussion at one point. If things need to be handled differently in the end, I suggest creating an new issue in JIRA or a new PR to discuss things again.

PR #23, opened 2020-04-24
Static linking support for non windows systems

stephen-webb:
This PR implements the documented behavior of the cmake options APR_STATIC=yes, APU_STATIC=yes and BUILD_SHARED_LIBS=no for non-windows systems.

Using the APR_STATIC=yes (not the default) and APU_STATIC=yes (not the default) and BUILD_SHARED_LIBS=yes (the default) builds log4cxx as a dynamically shared object (DSO) that includes APR and APR-Util in the DSO. 

Including APR and APR-Util in the log4cxx DSO (or including log4cxx in the application) simplifies application deployment.


src/cmake/FindAPR-Util.cmake
```diff
@@ -34,7 +34,14 @@ find_program(APR_UTIL_CONFIG_EXECUTABLE
 mark_as_advanced(APR_UTIL_CONFIG_EXECUTABLE)
 if(EXISTS ${APR_UTIL_CONFIG_EXECUTABLE})
     _apu_invoke(APR_UTIL_INCLUDE_DIR   --includedir)
-    _apu_invoke(APR_UTIL_LIBRARIES     --link-ld)
+    if (APU_STATIC OR NOT BUILD_SHARED_LIBS)
+      _apr_invoke(_apr_util_link_args  --link-ld)
```
ams-tschoening, line 38: Is `_apr_invoke` really correct here when everything else uses `_apu_invoke`?

src/cmake/FindAPR-Util.cmake
```diff
@@ -48,5 +55,5 @@ else()
     endif()
 endif()
 
-find_package_handle_standard_args(APR-Util DEFAULT_MSG
+find_package_handle_standard_args(APR-Util
```
ams-tschoening, line 59: How is removing `DEFAULT_MSG` related to this PR?

src/cmake/FindAPR-Util.cmake
```diff
@@ -34,7 +34,14 @@ find_program(APR_UTIL_CONFIG_EXECUTABLE
 mark_as_advanced(APR_UTIL_CONFIG_EXECUTABLE)
 if(EXISTS ${APR_UTIL_CONFIG_EXECUTABLE})
     _apu_invoke(APR_UTIL_INCLUDE_DIR   --includedir)
-    _apu_invoke(APR_UTIL_LIBRARIES     --link-ld)
+    if (APU_STATIC OR NOT BUILD_SHARED_LIBS)
+      _apr_invoke(_apr_util_link_args  --link-ld)
+      string(REGEX MATCH "-L([^ ]*)" _apr_util_L_flag ${_apr_util_link_args})
```
ams-tschoening, line 39: `-L([^ ]*)` does allow `-L` without any path. Is that really what you want, e.g. because the current working dir is used then? Or wouldn't `-L([^ ]+)` or `-L([^\s]+)` be better? The latter has less problems with newlines, if those can occur at all.

src/cmake/FindAPR.cmake
```diff
@@ -33,7 +33,14 @@ find_program(APR_CONFIG_EXECUTABLE
 mark_as_advanced(APR_CONFIG_EXECUTABLE)
 if(EXISTS ${APR_CONFIG_EXECUTABLE})
     _apr_invoke(APR_INCLUDE_DIR  --includedir)
-    _apr_invoke(APR_LIBRARIES  --link-ld)
+    if (APR_STATIC OR NOT BUILD_SHARED_LIBS)
+      _apr_invoke(_apr_link_args  --link-ld)
+      string(REGEX MATCH "-L([^ ]*)" _apr_L_flag ${_apr_link_args})
```
ams-tschoening, line 38: Same like above regarding the regular expression.

src/cmake/FindAPR-Util.cmake
stephen-webb: Opps -Sould all be 'apu' - I will fix those typos

src/cmake/FindAPR-Util.cmake
stephen-webb: Just so you can see the library being used - now the message output is consistent with FindAPR

src/cmake/FindAPR.cmake
stephen-webb: The regex is designed to work with the output generated by the 'apr-1-config' installed by APR install.

If** the -L and the path are separated, the logic would need to find the argument after '-L'.

src/cmake/FindAPR-Util.cmake
stephen-webb: The regex is designed to work with the output generated by the 'apu-1-config' installed by the APR-Util install.

I understand the Apache runtime will be merging APR-Util into APR in their next release so this script does not need to handle any future changes.

ams-tschoening: Is it necessary that you match `-L` without any further path? If not, please change to `-L([^\s]+)`, which matches available paths only. That seems far less surprising, `*` made me wonder about the purpose.

If you really need to match `-L` without any path, in my opinion this is something which should be explicitly documented. I wouldn't expect that and would be interested in the reason then.

stephen-webb: The path provided by apu-1-config is needed so cmake "find_library" function can locate libaprutil-1.a on Linux systems which support multiple binary interfaces and put archives in a sub-directory of /lib and /usr/lib (e.g. arm-linux-gnueabihf). The captured path is provided as a hint to find_library. CMake can also be told in which sub-directory to look in by setting CMAKE_FIND_LIBRARY_CUSTOM_LIB_SUFFIX.

If the captured path is empty, no hint will be provided to find_library. If cmake does not find the library, the "find_package_handle_standard_args" function will report that error.

The regex does not try to match "-L foo/bar" as that is not the form produced by apu-1-config.

I will change the regex to "-L[^ ]+" to improve clarity. The perl '\s' escape sequence is not recognised by cmake.

PR #24, opened 2020-06-02
Implement mapfilter

pm-cfs:
# Summary
This PR implements the mapfilter for log4cxx.

# Changes
Added implementation of mapfilter

_NOTES:_ the original mapfilter in log4j is a plugin; therefore, it uses a different set of tags entirely. The big question is would we rather copy that implementation syntax even though it would mean much deeper changes to log4cxx, or do we just want the functionality. I went with functionality as this was a pretty trivial addition and uses the existing codebase. My question on the mailing list had one agreeing response and no disagreements.

Using the existing filter syntax, we now have:

	<filter class="MapFilter">
		<param name="user.ip" value="127.0.0.1" />
		<param name="user.name" value="test2" />
		<param name="Operator" value="AND" />
		<param name="AcceptOnMatch" value="false" />
	</filter>

# Testing
There is an addition to the filtertests unit test which exercises the mapfilter logic.
You can also add a customized version of the above tag into your logging system and test in-place as we have done.

### DIFF
--- a/src/main/include/log4cxx/filter/mapfilter.h
+++ b/src/main/include/log4cxx/filter/mapfilter.h
@@ -32,14 +32,14 @@ namespace filter
  * A Filter that operates on a Map.
  */
 
- using keyvallist = std::map<LogString, LogString>;
+using KeyValList = std::map<LogString, LogString>;
 
 class LOG4CXX_EXPORT MapFilter: public log4cxx::spi::Filter
 {
 	private:
-		bool acceptOnMatch{true};
-		bool mustMatchAll{false};	// true = AND; false = OR
-		keyvallist mapKeyValList;
+		bool acceptOnMatch = true;
+		bool mustMatchAll = false;  // true = AND; false = OR
+		KeyValList mapKeyValList;
 
 	public:
 		DECLARE_LOG4CXX_OBJECT(MapFilter)
@@ -48,7 +48,7 @@ class LOG4CXX_EXPORT MapFilter: public log4cxx::spi::Filter
 		LOG4CXX_CAST_ENTRY_CHAIN(log4cxx::spi::Filter)
 		END_LOG4CXX_CAST_MAP()
 
-		MapFilter(){}
+		MapFilter() {}
 
 		/**
 		Set options
--- a/src/main/include/log4cxx/filter/mapfilter.h
+++ b/src/main/include/log4cxx/filter/mapfilter.h
@@ -19,26 +19,87 @@
 
 #include <log4cxx/spi/filter.h>
 
+#if defined(_MSC_VER)
+	#pragma warning ( push )
+	#pragma warning ( disable: 4251 )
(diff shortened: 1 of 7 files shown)


ams-tschoening:
Thanks for the PR, just for reference, this is the thread referred to:

http://mail-archives.apache.org/mod_mbox/logging-dev/202005.mbox/%3CCAAWxiMAKcN1iM0Aj_g0U3-WHMyPVSbZ9_ph9uLGP4o9bvVUfsw%40mail.gmail.com%3E

src/main/include/log4cxx/filter/mapfilter.h
```diff
@@ -19,26 +19,87 @@
...
+
 namespace log4cxx
 {
 namespace filter
 {
+/**
+ * A Filter that operates on a Map.
+ */
 
+ using keyvallist = std::map<LogString, LogString>;
```
ams-tschoening, line 35: That type reads somewhat badly, how about using `KeyValList` or `key_val_list` instead? I prefer the former, like most other custom types are named and that scheme is used by your variables as well.

src/main/include/log4cxx/filter/mapfilter.h
```diff
@@ -19,26 +19,87 @@
...
 
 class LOG4CXX_EXPORT MapFilter: public log4cxx::spi::Filter
 {
+	private:
+		bool acceptOnMatch{true};
```
ams-tschoening, line 40: C++11-features like this one need to be changed most likely to support legacy-compilers, simply because I'm still using one myself. :-/

http://docwiki.embarcadero.com/RADStudio/Rio/en/C%2B%2B11_Features_in_the_Classic_Compiler

src/main/include/log4cxx/filter/mapfilter.h
pm-cfs: done

src/main/include/log4cxx/filter/mapfilter.h
pm-cfs: I think this is complete; please let me know if your compiler has any other issues.

src/main/include/log4cxx/filter/mapfilter.h
```diff
@@ -19,26 +19,87 @@
...
+
 namespace log4cxx
 {
 namespace filter
 {
+/**
+ * A Filter that operates on a Map.
+ */
 
+using KeyValList = std::map<LogString, LogString>;
```
rm5248, line 35: Using 'using' inside of a header is generally bad practice; in this case though, while it may be okay I don't see the point of creating a new type(this is equivalent to `typedef KeyValList std::map<LogString,LogString>`).

pm-cfs: no problem; I switch over to how its done in the existing propertyfilter.h

PR #26, opened 2020-06-20
Fix missing const in method getThreshold

baldapps:
The method getThreshold wasn't const so it wasn't possible to call getThreshold on a const object and const cast was needed.


ams-tschoening:
Thanks for fixing! As this is a minor change, I won't update `changes.xml` for now.

PR #27, opened 2020-06-22
Fix missing const

baldapps:
Const was missing for method getDatePattern and a const_cast was needed if the object instance was const.


ams-tschoening:
Tests succeeded for me again, thanks!

PR #28, opened 2020-06-24
reorder find_package to build SO correctly

rm5248:
When building log4cxx with CMake, the BUILD_SHARED_LIBS option must be defined before finding APR and APR-util, as those use this variable.


ams-tschoening:
As I'm still [unable](issues.apache.org/jira/browse/LOGCXX-510) to build using CMAKE myself, I'll simply continue to trust that those changes make sense. Thanks for fixing.

PR #29, opened 2020-07-16
Prevent filenamepatterntestcase from failing

stephen-webb:
when the test is run in a time zone east of the prime meridian


ams-tschoening:
https://issues.apache.org/jira/projects/LOGCXX/issues/LOGCXX-273

ams-tschoening:
Tests still succeed for me, so am simply merging. Thanks for fixing!

PR #30, opened 2020-07-20
Documentation changes for release

stephen-webb:
This PR removes the maven generated Dependencies page from the web site.

The maven generated dependencies report is not appropriate for a C++ project.


INSTALL
```diff
@@ -15,24 +15,16 @@ limitations under the License.
...
+available in Git checkouts.
 
 autotools:
 site/building/autotools.html
 src/site/apt/building/autotools.apt
 
-Apache Ant:
-site/building/ant.html
-src/site/apt/building/ant.apt
+CMake:
```
ams-tschoening, line 24: How about using alphabetical order in this file? I've changed it at [the corresponding place](https://github.com/apache/logging-log4cxx/commit/96ef6546993a04e954bee4db4f0791a8138f3e00) as well.

INSTALL
```diff
@@ -15,24 +15,16 @@ limitations under the License.
...
+site/building/cmake.html
+src/site/apt/building/cmake.apt
 
 Apache Maven 2:
 site/building/maven.html
 src/site/apt/building/maven.apt
-
-Microsoft Visual Studio:
```
ams-tschoening, line 32: How about changing that to vcpkg? You did that [at other places](https://github.com/apache/logging-log4cxx/commit/e64b7cf0fbdd2f799cd83a7361d4ceefb75eff84#diff-7f6195436d46cda43818b2013775e6feR117) as well.

Please commit removing [vstudio.apt](https://github.com/apache/logging-log4cxx/blob/master/src/site/apt/building/vstudio.apt) as well.

pom.xml
```diff
@@ -372,7 +372,6 @@
 					<reportSet>
 						<reports>
 							<report>scm</report>
-							<report>dependencies</report>
```
ams-tschoening, line 375: Why do you want to remove [that page](https://logging.apache.org/log4cxx/next_stable/dependencies.html)? It currently somewhat documents that `log4j` is necessary for the tests, doesn't look too wrong to me. It would like to keep it for now.

src/site/xdoc/building/index.xml
```diff
@@ -54,10 +54,10 @@
 				<thead>
 					<tr>
 						<th id="os">OS</th>
-						<th id="os_x86_x64">x86 vs. x64</th>
+						<th id="os_x86_x64">Architecture</th>
```
ams-tschoening, line 57: Introducing a new header name is fine, but I suggest renaming those `*_x86_x64` phrases to `*_arch` as well. Things are clearer this way.

pom.xml
stephen-webb: I was trying to add the compile time dependencies (APR, APR-Util) and the testcase dependencies (sed, zip, gzip) to the report, but could not because the maven expects all entries in dependencies to be jar files.

I can only see that the dependencies can be mention elsewhere in documentation (hence the the building page change).

ams-tschoening: I get your point and tried the same in the past and failed as well. Tried again now and came up with some hacky approach:

	<dependency>
		<groupId>apr</groupId>
		<artifactId>util</artifactId>
		<version>1.5.4</version>
		<scope>system</scope>
		<systemPath>${user.home}/Documents/Svn/Src/Libs/trunk/C/X-OS/APR/apr-util/1.5.4/build/RAD 10.2/libs/libapr-util/Win32/Debug/out/libapr-util.lib</systemPath>
		<type>lib</type>
	</dependency>

That makes the build succeed and adds dependencies to the page, but still prints an error message and would be best to avoid any hard-coded path. I've [asked on SO](https://stackoverflow.com/questions/62995298/how-to-model-non-java-dependencies-not-available-in-any-repository-using-maven) about that problem, maybe there's some solution with keeping the plugin we are just not aware of.

ams-tschoening: According to comments on the SO-question I created the following result as an example:

Clipboard01

The problem is to find the binaries, I didn't find something suitable for ZIP and GZIP yet. GREP is only used as an additional example because it was easily available. In theory we could create our own `pom.xml` describing the individual missing dependencies and maintain those with the project. OTOH this might not be easier instead of simply writing text like you have done already, but that should be a bit more prominent than. I like the scope-stuff already provided by Maven, though.

Thoughts?

ams-tschoening:
The DIFF of my `pom.xml` for reference, was unable to attach that in the conversation directly.

[pom.xml.diff.txt](https://github.com/apache/logging-log4cxx/files/4952230/pom.xml.diff.txt)

ams-tschoening:
Closing here in favour of #35, that contains all the commits here already.

PR #31, opened 2020-08-04
Added documentation on non-english logging

rm5248:


ams-tschoening:
Is there anything special to do to make Doxygen use the MD-file or do things happen automatically when generating the site using Maven?

rm5248:
This is setup only for Maven at the moment.  Doxygen just needs to have markdown enabled to generate HTML pages from Markdown, and scan the correct directory for source files.

ams-tschoening:
Thinks worked OOB for me as well. I didn't like the top level menu entry, though, your text seems to be plain FAQ to me. So I replaced the available FAQ with a markdown document for Doxygen and added your text there. Makes it easier to spot the `setlocale` as well, because that is already mentioned.

The former FAQ generates the TOC automatically and looks a bit better, but maybe there's some way to get that with Doxygen as well at some point and the look can be improved using CSS in future as well. If introducing a Markdown-PoC, changing the FAQ seems to make more sense to me right now.

Thoughts?

Clipboard01

Clipboard02

rm5248:
The FAQ is probably the correct place to put it, but looking at it now it's really a sub-heading under 'Does log4cxx support unicode?', which also talks about the `setlocale`.  So I guess this is really an example of what happens if the locale is not set properly.

ams-tschoening:
Changed. Is this closing [LOGCXX-483](https://issues.apache.org/jira/browse/LOGCXX-483) as well? Then I would mention this in `changes.xml` as well.

I have the feeling setting the locale is only a workaround. Looking at the last comment in the issue, do you remember under which circumstances which decode/encoder has been provided? Does `setlocale(LC_ALL, "")` result in a decode for ANSI_X3.4-1968 or UTF-8?

rm5248:
This is related to LOG4CXX-483.  I don't think there's really a bug with LOG4CXX-483, but this should help clarify.

According to the [libc documentation](https://www.gnu.org/software/libc/manual/html_node/Setting-the-Locale.html), all programs start in the "C" locale by default.  The "C" locale is the [same as ANSI_X3.4-1968](https://stackoverflow.com/questions/48743106/whats-ansi-x3-4-1968-encoding).  So before you do `setlocale`, the value that I get is ANSI_X3.4-1968, which results in the output being all question marks as the log text provided is invalid ASCII.

To quote the glibc documentation:

> To use the locales specified by the environment, you must call setlocale. 

which explains why it works after you call `setlocale`, as my locale is en_US.UTF-8.

See also [this SO poston setting the default locale in C++](https://stackoverflow.com/questions/571359/how-do-i-set-the-proper-initial-locale-for-a-c-program-on-windows).  The potentially important thing is that you may need to actually do the following:

```
std::setlocale( LC_ALL, "" ); /* Set locale for C functions */
std::locale::global(std::locale("")); /* set locale for C++ functions */
```

ams-tschoening:
> According to the libc documentation, all programs start in the "C" locale by default. The "C" locale is the same as ANSI_X3.4-1968. So before you do setlocale, the value that I get is ANSI_X3.4-1968, which results in the output being all question marks as the log text provided is invalid ASCII.

That makes sense, I'll add those details to the FAQ and close the JIRA-issue mentioning these things as well.

PR #34, opened 2020-08-08
Fixed an infinite loop in "Transcoder::encode"

ams-tschoening:
Transcoder::encode(const LogString&, std::wstring&)" by appending LOSSCHAR. The most likely alternative would be to throw on any invalid input instead, but that's not the case right now and LOSSCHAR used in other places as well.

https://issues.apache.org/jira/browse/LOGCXX-398


rm5248:
Looks good to me - I went and added a test case as well to validate that it works correctly, if you want to take a look.

PR #35, opened 2020-08-08
Changed some docs and model dependencies for tests etc. using MVN.

ams-tschoening:
This is an improved version o #30, which DOES NOT remove the page about dependencies. Reason is that without that page, one needs to model the dependencies otherwise and that is not as easy at it sounds:

Currently, APR etc. are mostly only mentioned in texts, while the deps-plugin of MVN can maintain those in a table, with concrete version number etc. Additionally, it's difficult to distinguish in the texts between what is necessary to build the lib vs. at runtime vs. for test only. `sed` e.g. seems to be only necessary for tests, while `gzip` seems to NOT be necessary to build, but at runtime, because it is used by `gzcompressaction.cpp`. Therefore it's as well necessary for tests, but not only for tests. `log4j` OTOH is necessary for tests as well and even a plain Java-resource available in public repos.

All those difference can be modeled using the MVN-deps plugin in theory, so this is providing a PoC to discuss about. All commits from #30 are merged here already, so #30 might simply be closed in favour of this discussion. Even if the result is to NOT use the MVN-deps, that approach could simply be removed here and the parts of #30 merged anyway.

I've uploaded the changed site using this branch to make discussion easier:

https://logging.apache.org/log4cxx/next_stable/dependencies.html


rm5248:
If we do want to move away from maven in the not-too-distant future, something like this may be more appropriate for the dependencies: https://github.com/apache/logging-log4cxx/tree/dependencies-markdown
It looks like this:
Screenshot from 2020-08-08 13-02-29


The maven dependency plugin is more geared towards Java code, since it can pull in dependencies and transitive dependencies of maven projects.

ams-tschoening:
I don't like those markdown-tables too much, seem pretty difficult to maintain in my opinion. But I'm fine with that approach as well. The following sentence makes me wonder, though:

> Note that the correct binary for log4j will be downloaded and used automatically.

Is that really the case? I needed to maintain that JAR on my own and need to issue Java compiling in a custom script executing all tests as well:

	set log4j_CLASSPATH=%TEST_PROJ%\log4j-1.2.17.jar
	set SOCKET_SERVER_SOURCES=%TEST_SRC%\java\org\apache\log4j\net\ShortSocketServer.java
	set SOCKET_SERVER_CLASSPATH=%TEST_RES%;%log4j_CLASSPATH%
	set SOCKET_SERVER_COMMAND="C:\Program Files\Java\jdk-8\bin\java.exe" -classpath "%SOCKET_SERVER_CLASSPATH%" org.apache.log4j.net.ShortSocketServer 8 input/socketServer

	pushd "%TEST_RES%"
	javac -d "." -classpath "%log4j_CLASSPATH%" "%SOCKET_SERVER_SOURCES%"
	"%TEST_PROJ%\Win32\Debug\out\%~n0.exe" %*
	popd

[test_all.cmd.txt](https://github.com/apache/logging-log4cxx/files/5049958/test_all.cmd.txt)

rm5248:
> I don't like those markdown-tables too much, seem pretty difficult to maintain in my opinion. But I'm fine with that approach as well. The following sentence makes me wonder, though:
> 
> > Note that the correct binary for log4j will be downloaded and used automatically.
> 
> Is that really the case? I needed to maintain that JAR on my own and need to issue Java compiling in a custom script executing all tests as well:

That works for me just fine, although maybe its an issue with your setup?  The JAR will get downloaded [here](https://github.com/apache/logging-log4cxx/blob/afcad955e73c564517bc09e80ca9bebc663295d7/src/test/java/CMakeLists.txt#L2), although if [you go up one directory](https://github.com/apache/logging-log4cxx/blob/afcad955e73c564517bc09e80ca9bebc663295d7/src/test/CMakeLists.txt) it will only include the Java tests if the tools(e.g. javac) are found.  Perhaps it doesn't find them if they're not on your PATH?

ams-tschoening:
I'm simply not using CMAKE yet and thought you are talking about something MVN does automatically. I would suggest adding some hint that CMAKE only does those things automatically for people not using that, especially as we still provide autotools.

ams-tschoening:
I've added a hint, but leave merging to @rm5248 to not break his release-process.

rm5248:
Looks good to me - I'll merge it shortly once the release gets finalized.

rm5248 (approved):

PR #36, opened 2020-08-19
Remove MD5, add SHA-512 file extension.

ams-tschoening:
MD5 is not used anymore for releases and the SHA file extension needed to be different because of SHA-512.


ams-tschoening:
I didn't test things, it's more of a reminder that this needs to be done anyway and don't want to break release process of @rm5248.

PR #37, opened 2020-08-27
Search for programs needed for tests

rm5248:
Add a check for applications that are needed for tests to run successfully.


CMakeLists.txt
```diff
@@ -123,3 +123,10 @@ message(STATUS "  logchar type .................... : ${LOG4CXX_CHAR}")
 message(STATUS "  Using libESMTP .................. : ${HAS_LIBESMTP}")
 message(STATUS "  ODBC library .................... : ${HAS_ODBC}")
 message(STATUS "  syslog .......................... : ${HAS_SYSLOG}")
+
+if(BUILD_TESTING)
+message(STATUS "Applications required for tests:")
+message(STATUS "  zip ............................. : ${ZIP_EXE}")
```
ams-tschoening, line 129: The first thing I wondered when reading "*_EXE" was about Windows vs. non-Windows. How about changing the name to "*_BIN" or "*_APP" instead to make clear it's OS-independent?

ams-tschoening (changes requested):

CMakeLists.txt
rm5248: Sure, that makes sense.

ams-tschoening (approved):

PR #40, opened 2020-09-07
GitHub actions and OSX fixes

rm5248:
Add github actions to build automatically, as well as fix an error on OSX with regards to a stream destructor.


.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
ams-tschoening, line 1: Looking at the docs, this setting doesn't support the special value `input`, doesn't it?

Your problem was automatic conversion to CRLF, which was wrong, but keeping things as is would have prevented that problem as well. That's exactly what `autocrlf = input` does in the end and what I configure by default system wide. That's why things succeeded for me. LF stay LF, CRLF stay CRLF, LF will NEVER be CRLF unless one converts manually for some reason.

The downside of setting LF is that CRLF of the repo becomes LF in the working copy even if CRLF would be correct and needed to be changed manually, e.g. because tests would needed to be adopted etc. So if supported, `input` would be a better choice in my opinion. Just would like to make sure it isn't supported.

.gitignore
```diff
@@ -19,11 +19,15 @@ pom.xml.releaseBackup
 release.properties
 
 autom4te.cache/
+src/.vs
+src/**/__history
```
ams-tschoening, line 23: I have directories like the following as well in my tree:

> src/**/__recovery

Not sure if those are C++-Builder specific, as I use the same tree for Visual Studio and C++-Builder.

src/cmake/FindAPR-Util.cmake
```diff
@@ -29,7 +29,7 @@ endmacro(_apu_invoke)
 
 find_program(APR_UTIL_CONFIG_EXECUTABLE
     apu-1-config
-    PATHS /usr/local/bin    /usr/bin    C:/Progra~1/apr-util/bin
+    PATHS /usr/local/bin    /usr/bin    C:/Progra~1/apr-util/bin /usr/local/opt/apr-util/bin
```
ams-tschoening, line 32: I suggest grouping non-Windows paths after each other and put Windows last/first. Reads easier this way:

    /usr/local/bin /usr/local/opt/apr-util/bin /usr/bin C:/Progra~1/apr-util/bin 

Besides that, `C:/Progra~1/apr-util/bin` resolves differently depending on where the build is executed: On a 64 Bit system a 64 Bit path is used, while on a 32 Bit system a 32 Bit one is used, because that's the only existing one. Things get interesting if one wants to build 32 Bit on a 64 Bit system, in which case the 64 Bit path would be used in theory.

Is that intended or something to make more clear using an absolute path or simply to ignore for now or ...?

    07.09.2020  10:23    <DIR>          PROGRA~1     Program Files
    07.09.2020  10:23    <DIR>          PROGRA~2     Program Files (x86)

Apps normally deal with those problems by using environment variables like `%ProgramFiles%`. That resolves differently depending on the architecture of the current OS, the architecture of the current process etc.

src/cmake/FindAPR.cmake
```diff
@@ -28,7 +28,7 @@ endmacro(_apr_invoke)
 
 find_program(APR_CONFIG_EXECUTABLE
     apr-1-config
-    PATHS /usr/local/bin    /usr/bin    C:/Progra~1/apr/bin
+    PATHS /usr/local/bin    /usr/bin    C:/Progra~1/apr/bin /usr/local/opt/apr/bin
```
ams-tschoening, line 31: Group paths like for APU.

ams-tschoening (changes requested):
Minor suggestions to look at.

.gitattributes
rm5248: It doesn't look like it, but we could perhaps be a bit more specific.  I think we could do something like
```
src/test/witness/* -text
```
to tell git to not normalize the line endings, and then having a `* text=auto` entry for everything else(I haven't tried this out).  I don't view it as a big deal, since most(all?) IDEs should take care of that, although it's possible that we could get into a situation where \r\n is mixed with \n in the same file.

src/cmake/FindAPR-Util.cmake
rm5248: I'm not sure of the reason behind using Progra~1, that's something that Steven Webb added.  I assume it works if you have things installed globally on your system, but I don't think I've ever built it using the apr-config/apu-config executables.

Grouping the non-windows paths makes sense though, I can fix that easily.

.gitignore
rm5248: They may be; I pulled this from one of your branches, maybe it never got added to the gitignore?  It's not hard to add though.

.gitattributes
ams-tschoening: If we agree that `input` is not supported, keep things as they are. No need to make things unnecessary complex and `auto` pretty much always is a bad choice in my opinion. If someone needs to introduce CRLF in some test by purpose, this can be reconsidered.

.gitignore
ams-tschoening: If you pulled from me, than they are C++-Builder specific and you should add `__recovery` as well. I do have some `__astcache`-dirs as well, but not in the source tree. So let's focus on what is really needed. Don't know myself when `__recovery` gets created, though.

src/cmake/FindAPR-Util.cmake
ams-tschoening: @stephen-webb Should we change `C:/Progra~1/apr-util/bin` to become `%ProgramFiles%/apr-util/bin` or does it break for you? Using the environment variable instead of the concrete directory name looks more correct to me.

stephen-webb:
Yes, using the environment variable would be better, although there is not
any standard place to look for libraries in Windows AFAIK.

I copied those hints from another project without understanding the reason
they were chosen.

src/cmake/FindAPR-Util.cmake
ams-tschoening: I've successfully tested this PR with the following paths instead of the legacy short dir names and therefore suggest to change it:

    $ENV{ProgramFiles}/apr/bin
    $ENV{ProgramFiles}/apr-util/bin

ams-tschoening (approved):

PR #42, opened 2020-09-24
fix build without wchar

ffontaine:
Disable wencode without wchar or the build will fail on:

```
odbcappender.cpp: In static member function 'static void log4cxx::db::ODBCAppender::encode(wchar_t**, const LogString&, log4cxx::helpers::Pool&)':
odbcappender.cpp:362:22: error: 'wencode' is not a member of 'log4cxx::helpers::Transcoder'
  *dest = Transcoder::wencode(src, p);
                      ^~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/bab5329fdeb894471bfd5192ce04d3fbd2f9be5c

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/main/cpp/odbcappender.cpp
+++ b/src/main/cpp/odbcappender.cpp
@@ -357,10 +357,12 @@ void ODBCAppender::setSql(const LogString& s)
 	}
 }
 
+#if LOG4CXX_WCHAR_T_API || LOG4CXX_LOGCHAR_IS_WCHAR_T || defined(WIN32) || defined(_WIN32)
 void ODBCAppender::encode(wchar_t** dest, const LogString& src, Pool& p)
 {
 	*dest = Transcoder::wencode(src, p);
 }
+#endif
 
 void ODBCAppender::encode(unsigned short** dest,
 	const LogString& src, Pool& p)
--- a/src/main/include/log4cxx/db/odbcappender.h
+++ b/src/main/include/log4cxx/db/odbcappender.h
@@ -300,8 +300,10 @@ class LOG4CXX_EXPORT ODBCAppender : public AppenderSkeleton
 	private:
 		ODBCAppender(const ODBCAppender&);
 		ODBCAppender& operator=(const ODBCAppender&);
+#if LOG4CXX_WCHAR_T_API || LOG4CXX_LOGCHAR_IS_WCHAR_T || defined(WIN32) || defined(_WIN32)
 		static void encode(wchar_t** dest, const LogString& src,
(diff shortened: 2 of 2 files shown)


rm5248:
Shouldn't this also #ifdef out the function signature in the .h file?  Or would it make more sense to have the method implementation be just a stub?  I'm thinking that you could still create something that used the encode method, which would then fail at link-time instead of compile-time.

ffontaine:
I updated the PR to keep the `ODBCAppender::encode` function which won't do anything if `Transcoder::wencode` is not available.

ams-tschoening:
How does keeping `encode` as a no-op make sense here? Either it's called, then making it a no-op breaks on runtime only or it's not, then itself should be `#ifdef`ed in my opinion, not its implementation only.

ffontaine:
I updated the PR following the second suggestion of @rm5248 but perhaps I misunderstood it. I'll update the PR again to ifdef `ODBCAppender::encode`.

rm5248:
> How does keeping `encode` as a no-op make sense here? Either it's called, then making it a no-op breaks on runtime only or it's not, then itself should be `#ifdef`ed in my opinion, not its implementation only.

There's two ways to do this:
1.
```
// header file
#ifdef SOME_VARIABLE
void foo();
#endf

// cpp file
#ifdef SOME_VARIABLE
void foo(){
    // Some Implementation
    bar();
}
#endif
```

2.
```
// header file
void foo();

// cpp file
#ifdef SOME_VARIABLE
void foo(){
    // Some Implementation
    bar();
}
#else
void foo(){
  // No-op implemetation
}
#endif
```

The advantage of the first method is that if you don't have some feature, you can't use it in the code.  The advantage of the second example is that the function always exists, so that you can always call it from any code without having the calling code to check if it exists.  The second example is preferred for [Linux kernel coding](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#conditional-compilation) for this reason.  This does imply that you need to structure your code accordingly to make sure that there are no side-effects that you are expecting from the method you are calling, if it doesn't exist.

We have a (potential) problem with issues like this if you use objects such as the NTEventLogAppender.  I can include the file in Linux, but as soon as I try to compile an application that uses it I get a linker error, as it is Windows-only.  So if you are somehow using the NTEventLogAppender directly in a cross-platform application, you need to #ifdef out portions of your code instead of those operations being no-ops on Linux.

In this particular case, it seems like the first option is the best, since log4cxx may or may not have the `wchar_t` API.
-----

Otherwise, looks good to me.

PR #43, opened 2020-11-16
fix log4cxx mapfilter chaining

pm-cfs:
# Summary
Previously we either ACCEPTed or DENYed any log when this filter was present. This configuration prevented proper filter chaining.

# Changes
* Update mapfilter to return NEUTRAL when additional filters should be considered.
* Update tests to check for these values.

# Testing
Run the unit test.

To manually test: using any logging configuration where IP addreses are available in MDC (implementation dependent):
```
   <filter class="MapFilter">
      <param name="ip" value="192.169.0.2" />
      <param name="AcceptOnMatch" value="false" />
    </filter>
    <filter class="MapFilter">
      <param name="ip" value="192.169.0.3" />
      <param name="AcceptOnMatch" value="false" />
    </filter>
```
* In the first filter set the IP address to a machine/VM you want to ignore; entries from this machine won't be logged.
* In the second, change ip address to another machine to ignore.
* Now verify that the first and second are ignored and a third machine is accepted.

### DIFF
--- a/src/main/cpp/mapfilter.cpp
+++ b/src/main/cpp/mapfilter.cpp
@@ -82,10 +82,10 @@ Filter::FilterDecision MapFilter::decide(
 
 	if (acceptOnMatch)
 	{
-		return matched ? Filter::ACCEPT : Filter::DENY;
+		return matched ? Filter::ACCEPT : Filter::NEUTRAL;
 	}
 	else
 	{
-		return matched ? Filter::DENY : Filter::ACCEPT;
+		return matched ? Filter::DENY : Filter::NEUTRAL;
 	}
 }
--- a/src/test/cpp/filter/mapfiltertest.cpp
+++ b/src/test/cpp/filter/mapfiltertest.cpp
@@ -57,8 +57,8 @@ public:
 	}
 
 	/**
-	 * Check that MapFilter.decide() returns Filter.ACCEPT or Filter.DENY
-	 *   based on Accept on Match setting when key/value does not match
+	 * Check that MapFilter.decide() returns Filter.NEUTRAL
+	 *   when key/value does not match
 	 */
 	void test2()
 	{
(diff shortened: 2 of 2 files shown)


rm5248:
This looks good to me.  I'll merge shortly unless @ams-tschoening has any comments?

Unrelated note: would you be able to create some documentation on how to use the filters, since you use them?  I'm trying to expand the documentation that we do have, but that is a feature that I have never used before, so I don't know where to start.

pm-cfs:
Certainly. I was thinking about that too. I'll have a look at similar
documentation, but if you know of anything similar off the top of your
head, please feel free to direct me in an appropriate direction.

-Pat



Patrick Mills • Principal Software Engineer
Cerberus, LLC • www.cerberusftp <http://www.cerberusftp.com/>.com
<http://www.cerberusftp.com/> • patrick.mills@cerberusftp.com
<alissa.mcguire@cerberusftp.com>
LinkedIn <https://www.linkedin.com/company/cerberus-ftp/> • Facebook
<https://www.facebook.com/cerberusftp> • Twitter
<https://twitter.com/cerberusftp> • Get Support
<https://www.cerberusftp.com/support/> • Request A Quote
<https://www.cerberusftp.com/contact/request-a-quote/>

rm5248:
The Log4j2 documentation would probably be the best place to start: http://logging.apache.org/log4j/2.x/manual/filters.html

pm-cfs:
I wrote up some MapFilter documentation based on the Log4j2 docs
formatting. I have it in Google Docs and it exports to PDF ok, but doesn't
do very well going to Word / RTF. I could also try to export to HTML.

What's the best way to deliver / share it? Do we want to create a docs
folder and submit as a PR? Will there be a standard template or CMS to
enter it into?

>

ams-tschoening:
Have a look at [src\site\markdown](https://github.com/apache/logging-log4cxx/tree/master/src/site/markdown), some `filters.md` there might be a good idea. OTOH, depending on how much you have actually written, providing your docs as APIDOC to the concrete classes might make sense as well.

PR #45, opened 2020-12-02
LOG4CXX-513 Reordered some rolling code to create intermediate direct…

rm5248:
…ories

### DIFF
--- a/src/main/cpp/rollingfileappender.cpp
+++ b/src/main/cpp/rollingfileappender.cpp
@@ -340,11 +340,13 @@ bool RollingFileAppenderSkeleton::rollover(Pool& p)
 						}
 						else
 						{
+							closeWriter();
+							setFile(rollover1->getActiveFileName());
+							// Call activateOptions to create any intermediate directories(if required)
+							FileAppender::activateOptions(p);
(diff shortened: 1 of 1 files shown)


src/main/cpp/rollingfileappender.cpp
```diff
@@ -340,11 +340,13 @@ bool RollingFileAppenderSkeleton::rollover(Pool& p)
 						}
 						else
 						{
+							closeWriter();
```
ams-tschoening, line 343: [activateOptions](https://github.com/apache/logging-log4cxx/blob/9cf215f9721aa6add98674bfc706452e62268b50/src/main/cpp/fileappender.cpp#L155) calls [setFile](https://github.com/apache/logging-log4cxx/blob/9cf215f9721aa6add98674bfc706452e62268b50/src/main/cpp/fileappender.cpp#L164) as well which calls [closeWriter](https://github.com/apache/logging-log4cxx/blob/9cf215f9721aa6add98674bfc706452e62268b50/src/main/cpp/fileappender.cpp#L281) already. So the explicit call might not be necessary anymore.

Though, just a hint, not worth blocking the PR.

ams-tschoening (approved):

PR #46, opened 2020-12-26
LOGCXX-324 Split up large syslog messages

rm5248:


src/main/include/log4cxx/net/syslogappender.h
```diff
@@ -136,6 +148,16 @@ class LOG4CXX_EXPORT SyslogAppender : public AppenderSkeleton
 			return facilityPrinting;
 		}
 
+		inline void setMaxMessagelength(int maxMessageLength1)
```
ams-tschoening, line 151: * Lower case `l` seems wrong: `setMaxMessagelength` vs. `setMaxMessageLength`, like `getMaxMessageLength`.

* `maxMessageLength1` to not accidently use the instance variable?

* How about a shorter property like `maxMsgLength` instead? I can easily live with yours of course, sounds a bit noisy in my opinion, though.

ams-tschoening (changes requested):

src/main/include/log4cxx/net/syslogappender.h
rm5248: * CamelCase fixed
* maxMessageLength1 is a similar pattern to what is used in other places(see facilityprinting a few lines above).  I don't really like it either, but I don't want to have different coding styles if we can avoid it.
* Personally, I prefer being a bit more verbose(message vs msg), especially when it's in a public API.

ams-tschoening (approved):

PR #50, opened 2021-01-04
LOGCXX-510: Additional errors @ TSCHOENING

ams-tschoening:
Building the current MASTER, I came across additional errors for [LOGCXX-510](https://issues.apache.org/jira/browse/LOGCXX-510?filter=-2). Would be great if someone could have a look at and tell me if those changes make any sense, if they break something, if my setup is broken etc. The only thing I did compared to a few months ago I remember is updating VS 19 community regularly.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,10 +1,11 @@
+cmake_policy(SET CMP0079 NEW)
+
 add_subdirectory(main)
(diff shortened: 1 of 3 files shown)


rm5248:
Hmm... it was broken for you, but the automatic builds on github worked?  That seems very weird.  I'm going to try it out myself now.

This was one of the runs the other day that worked fine: https://github.com/apache/logging-log4cxx/runs/1622829005

PR #51, opened 2021-01-31
Old build tools removal

rm5248:
Remove the old build tools(ant, maven, autotools) as these have at this point been superseded by the cmake/doxygen process that we merged in #49


rm5248:
I'll merge shortly if there are no comments.

ams-tschoening:
I couldn't find any mail with your requested review, but mails have changed for me, so they might have been dropped for some reason. Anyway, the only thing I've recognized are the `release_*`-scripts. Some of those are pretty coupled with MVN. Do you want to keep those and change as necessary later or ...?

rm5248:
No, I actually forgot to remove those.  Those have now been removed.

ams-tschoening (approved):

PR #52, opened 2021-02-15
Log4cxx Qt support

rm5248:
Add a message handler function that is used for Qt so that log messages from Qt can be routed through log4cxx.

Note that this is built as a separate library - this is intentional, as many people may not be using Qt, so building it into the main log4cxx library doesn't make much sense.  This is so that (for example) Debian could build a separate package, log4cxx-qt, that is installed along with the main log4cxx .so files.


ams-tschoening (approved):
I don't use QT, so only tested things with my old setup and didn't ran into problems.

> 100% tests passed, 0 tests failed out of 61
> Total Test time (real) =  98.06 sec

PR #53, opened 2021-02-18
Smart pointer support

rm5248:
This PR should switch log4cxx over to using smart pointers(either std::shared_ptr or boost:shared_ptr) as appropriate.  It also uses synchronization as defined by the C++ standard(std::mutex) or boost as appropriate.

I believe that this will fix the following issues:
LOGCXX-486
LOGCXX-517(discovered while doing this conversion)
LOGCXX-485
LOGCXX-322
LOGCXX-430


src/main/cpp/domconfigurator.cpp
```diff
@@ -331,10 +321,12 @@ void DOMConfigurator::parseErrorHandler(Pool& p,
 	AppenderMap& appenders)
 {
 
-	ErrorHandlerPtr eh = OptionConverter::instantiateByClassName(
+	ErrorHandlerPtr eh;
+	std::shared_ptr<Object> obj = OptionConverter::instantiateByClassName(
```
ams-tschoening, line 325: There are multiple places using `std::shared_ptr` like above, shouldn't that be `log4cxx::` instead?

src/cmake/boost-fallback/boost-std-configuration.h.cmake
```diff
@@ -0,0 +1,94 @@
...
+}
+#endif
+
+#if STD_MUTEX_FOUND
+#include <mutex>
+#include <condition_variable>
+namespace ${NAMESPACE_ALIAS} {
+    typedef std::mutex mutex;
+    template <typename T>
+    using unique_lock = std::unique_lock<T>;
```
ams-tschoening, line 33: This syntax doesn't work for C++ less than 11, I'm running into this with my legacy C++-Builder 10.2 UPD 3 with 32 Bit BCC. So simply don't support those old compilers anymore? On the mailing list we had a discussion in the past about different versions of the lib or not etc.

https://stackoverflow.com/a/2795024/2055163

src/cmake/boost-fallback/boost-std-configuration.h.cmake
```diff
@@ -0,0 +1,94 @@
...
+#endif
+
+#if STD_ATOMIC_FOUND
+#include <atomic>
+namespace ${NAMESPACE_ALIAS} {
+    template <typename T>
+    using atomic = std::atomic<T>;
+}
+#elif Boost_ATOMIC_FOUND
+#include <boost/atomic.hpp>
```
ams-tschoening, line 87: Another example for problems with too old compilers: My C++-Builder 10.2 UPD 3 with 32 Bit BCC provides boost 1.39 in theory, which seems to use a [different namespace for atomic](https://www.boost.org/doc/libs/1_39_0/boost/interprocess/detail/atomic.hpp). Additionally, exactly those parts of boost seem to not be available for my compiler at all, I simply can't find associated files. :-) when searching for `atomic.hpp`, I onl y get results named like `atomic_count_*.hpp` from `boost\smart_ptr\detail`.

src/cmake/boost-fallback/boost-std-configuration.h.cmake
ams-tschoening: Using inheritance compiles, but one needs to provide all CTORs, operators etc. manually this way:

```cpp
template <typename T>
class unique_lock : public boost::unique_lock<T>
{
	public:
		unique_lock(const T& mutex) : boost::unique_lock(mutex) {}
};
```

src/cmake/boost-fallback/boost-std-configuration.h.cmake
ams-tschoening: [Only at least C++11 will be supported](https://lists.apache.org/thread.html/re744feb89b0bf9a15b03d145a88fe0d0be48ad069d12a31b8ba5d461%40%3Cdev.logging.apache.org%3E) in future and especially `BCC32` is a dead-end for everyone still using it anyway. Embarcadero prefers its CLANG-based alternatives as well for some years now, people need to upgrade their projects.

src/cmake/boost-fallback/boost-std-configuration.h.cmake
ams-tschoening: [Only at least C++11 will be supported](https://lists.apache.org/thread.html/re744feb89b0bf9a15b03d145a88fe0d0be48ad069d12a31b8ba5d461%40%3Cdev.logging.apache.org%3E) in future and especially `BCC32` is a dead-end for everyone still using it anyway. Embarcadero prefers its CLANG-based alternatives as well for some years now, people need to upgrade their projects.

src/main/cpp/domconfigurator.cpp
ams-tschoening: [Only at least C++11 will be supported](https://lists.apache.org/thread.html/re744feb89b0bf9a15b03d145a88fe0d0be48ad069d12a31b8ba5d461%40%3Cdev.logging.apache.org%3E) in future, so it doesn't make too much sense anymore to migrate those calls to `log4cxx::`. Instead, the project is simply going to depend on [std::shared_ptr without BOOST-fallback](https://lists.apache.org/thread.html/r80a4dbbb719684c899223a1540e3d608c8fa74928dd9adfe46a5ae06%40%3Cdev.logging.apache.org%3E), so the above code can either stay as-is or might be changed to use already available macros `LOG4CXX_PTR_DEF`.

ams-tschoening (approved):
All tests passed with my VS 2019 16.9.0, so feel free to merge.

PR #54, opened 2021-03-12
[Logcxx-515] Add fmt support

rm5248:
Added macros and documentation no how to utilize [fmt](https://fmt.dev/latest/index.html) with log4cxx.

This intentionally does not utilize fmt within log4cxx, to avoid dependencies that are not needed if your application does not utilize fmt.  This has the drawback of not always being available, and there are no checks(apart from trying to compile) that would tell you this.

Utilizing fmt is also slightly faster than doing `operator<<` on all of the arguments to a log message.

Possible concerns:

1. Does the macro name (just adding _FMT) make sense?
2. Do we also need to provide macros for localized log macros(LOG4CXX_L7DLOG1...)?  I have never used these macros, so I didn't include them.

### DIFF
--- a/src/examples/cpp/CMakeLists.txt
+++ b/src/examples/cpp/CMakeLists.txt
@@ -10,3 +10,12 @@ endforeach()
 configure_file( custom-appender.xml
     ${CMAKE_CURRENT_BINARY_DIR}/custom-appender.xml
     COPYONLY )
+
+# Custom handling for format string example, since it utilizes libfmt
+find_package(fmt)
+if(${fmt_FOUND})
+    add_executable( format-string format-string.cpp )
+    target_compile_definitions(format-string PRIVATE ${LOG4CXX_COMPILE_DEFINITIONS} ${APR_COMPILE_DEFINITIONS} ${APR_UTIL_COMPILE_DEFINITIONS} )
+    target_include_directories(format-string PRIVATE ${CMAKE_CURRENT_LIST_DIR} $<TARGET_PROPERTY:log4cxx,INCLUDE_DIRECTORIES>)
+    target_link_libraries(format-string PRIVATE log4cxx ${APR_UTIL_LIBRARIES} ${XMLLIB_LIBRARIES} ${APR_LIBRARIES} ${APR_SYSTEM_LIBS} fmt::fmt)
+endif(${fmt_FOUND})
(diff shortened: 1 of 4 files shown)


ams-tschoening:
I get the following warning in Visual Studio when building everything:

```
CMake Warning at src/examples/cpp/CMakeLists.txt:15 (find_package):
  By not providing "Findfmt.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "fmt", but
  CMake did not find one.

  Could not find a package configuration file provided by "fmt" with any of
  the following names:

    fmtConfig.cmake
    fmt-config.cmake

  Add the installation prefix of "fmt" to CMAKE_PREFIX_PATH or set "fmt_DIR"
  to a directory containing one of the above files.  If "fmt" provides a
  separate development package or SDK, be sure it has been installed.
```

As I'm somewhat new to CMAKE, that made me wonder. Does it make sense to document that such a warning can simply be ignored in the CMAKE-file and/or the newly added docs?  Something like the following?

```
# Custom handling for format string example, since it utilizes "libfmt". Without that lib being
# provided, a warning is logged by CMAKE, which can easily be ignored. Things are NOT build without
# that lib and CMAKE is simply telling one that.
```

And/or:

```
A full example can be seen in the "src/examples/cpp/format-string.cpp" file. Attention: Without 
"libfmt" being available, CMAKE might log a warning telling one exactly that and NOT building the
example. So that warning can simply be ignored.
```

Or should we make looking for that package simply [silent](https://cmake.org/cmake/help/latest/command/find_package.html)?

```
find_package(fmt QUIET)
```

rm5248:
> Or should we make looking for that package simply [silent](https://cmake.org/cmake/help/latest/command/find_package.html)?
> 
> ```
> find_package(fmt QUIET)
> ```

I prefer to make it quiet, since it's 1. only an example, and 2. at the end of configuration, we print out some stats/configuration information anyway, where we can print out if we're using it or not.  Number 2 may be more useful later, as when I get around to [LOGCXX-514](https://issues.apache.org/jira/browse/LOGCXX-514) a more verbose error message would be useful(I'm planning on it being an optional dependency).

ams-tschoening (approved):
Sounds reasonable to me as well.

PR #55, opened 2021-03-13
[LOGCXX-322] Fix crashing on exit from multithreaded application

rm5248:
By turning the `LogManager::repositorySelector` into a std::shared_ptr as part of the smart pointer conversion, it is now possible to know when the `LogManager::repositorySelector` is valid or not by configuring the loggers to use a weak reference to the `RepositorySelector` instead of a raw pointer.

This also adds a test that runs multiple threads, and a random thread will then call `std::exit` to force a (semi-abnormal) shutdown.

It is possible that log messages are lost because of this, but fixing that probably requires some more invasive changes.


ams-tschoening (approved):
Things look good and worked for me as well.

PR #60, opened 2021-04-02
Turn the threadname into a thread_local variable

rm5248:
Every time a LoggingEvent is created, part of the information that is stored in the LoggingEvent is the name of the thread that the log is from.  By turning this into a `thread_local` variable and only setting it once, I was able to to get a speedup of 25-40% depending on the configuration that I'm using.

Note that since thread_local is C++11, the only check we do to see if it's supported or not is to determine if the compiler is Borland or not - everybody else seems to be just fine with it(https://en.cppreference.com/w/cpp/compiler_support).  We could add a more thorough check if needed.

### DIFF
--- a/src/main/cpp/loggingevent.cpp
+++ b/src/main/cpp/loggingevent.cpp
@@ -222,28 +222,35 @@ LoggingEvent::KeySet LoggingEvent::getPropertyKeySet() const
 const LogString LoggingEvent::getCurrentThreadName()
 {
 #if APR_HAS_THREADS
+#if LOG4CXX_THREAD_LOCAL
+	thread_local LogString thread_name;
+	if( thread_name.size() ){
+		return thread_name;
+	}
+#endif /* LOG4CXX_THREAD_LOCAL */
+
 #if defined(_WIN32)
 	char result[20];
 	DWORD threadId = GetCurrentThreadId();
 	apr_snprintf(result, sizeof(result), LOG4CXX_WIN32_THREAD_FMTSPEC, threadId);
 #else
-	thread_local LogString thread_name;
-	if( thread_name.size() ){
-		return thread_name;
-	}
-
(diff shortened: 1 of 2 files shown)


ams-tschoening (changes requested):
I would like to suggest a slightly different implementation reducing the number of pre-processor checks in favour of the new macro to directly (un/)define `thread_local` as attribute. A variable of type `LogString` is necessary anyway and implicitly created by `LOG4CXX_DECODE_CHAR` currently, which could be avoided by using `thread_name` always.

[different_impl.diff.txt](https://github.com/apache/logging-log4cxx/files/6263238/different_impl.diff.txt)

With a different signature of `decode`, things could become even simpler in the long term:

```cpp
	static void decode(const std::string& src, LogString& dst);

	log4cxx::helpers::Transcoder::decode((const char*) result, thread_name);
	return thread_name;
```

vs.

```cpp
	static LogString& decode(const std::string& src, LogString& dst);

	return log4cxx::helpers::Transcoder::decode((const char*) result, thread_name);
```

Thoughts?

rm5248:
That is actually a much simpler implementation.  I didn't think about doing it that way, as I was thinking about having a fallback method of having a map of thread<->name.  That's a potential future optimization for environments that don't have thread_local.

Changing the return value of the `Transcoder::decode` may make sense in the future - passing the string to modify into the function looks like an optimization that may no longer be relevant with newer C++ versions, so I would have to do some profiling(which is how I found this in the first place).

ams-tschoening (approved):

ams-tschoening:
> Changing the return value of the Transcoder::decode may make sense in the future - passing the string to modify into the function looks like an optimization that may no longer be relevant with newer C++ versions[...]

Keep in mind that the current implementation really APPENDS to the given string instead of returning a new value only. That behaviour would need to be kept most likely anyway.

PR #61, opened 2021-04-13
MapFilter documentation in Markdown

pm-cfs:
Documentation for MapFilter was requested, but I couldn't reconcile the various format/location suggestions, so I'm providing this in Markdown which should be straightforward to convert to whatever final format is decided. It seemed better to have the documentation than not and I didn't want to accidentally lose it.

### DIFF
--- a/src/site/markdown/filters/MapFilter.md
+++ b/src/site/markdown/filters/MapFilter.md
@@ -1,0 +1,81 @@
+# MapFilter
+
+The MapFilter allows filtering against data elements that are in the Mapped Diagnostic Context (MDC).
+
+| **Parameter Name** 	| **Type**  	| **Description**                                                                                                                                                                                                    	|
+|:-------------------	|:----------	|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------	|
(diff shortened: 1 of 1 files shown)


ams-tschoening (changes requested):
I would prefer backticks instead of italic for "code", which would avoid using `_[...]_` and `*[...]*` as well. Additionally and unless done for a special reason, I would prefer a code block like the following for the XML-examples as well. This should even keep newlines when rendered, so `<br/>` could be removed?

    ```xml
    [...]
    ```

We should keep in mind that code and XML have a semantic meaning and should be declared as such in Markdown, not only visibly formatted.

pm-cfs:
Requested changes made; my markdown engine doesn't do a great job with this, but that's my problem. I understand the semantics request and hopefully this works better for everyone - viewing & copy/paste.

ams-tschoening (approved):
Didn't know I could implement changes on my own, so have done that now. Please review for you again, if you can live with my changes.

pm-cfs:
I'm ok with these; I would note that the tabs are a visibility thing for non-markdown and/or compatibility for some renderers that use them.
I didn't notice that my editor removed the indentions; that's definitely good for everyone.

rm5248 (approved):

PR #63, opened 2021-04-17
Use new LOG4CXX_SOURCE_DIR in the event that this is not the top-leve…

rm5248:
…l cmake file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,11 @@ include(src/cmake/projectVersionDetails.cmake)
 project(log4cxx VERSION ${log4cxx_VER} LANGUAGES CXX)
 include(CTest)
 
+# If you are including log4cxx from a higher-level CMake file(perhaps as a submodule?)
+# CMAKE_SOURCE_DIR will refer to the first CMakeLists.txt.  Since we need some files
+# in relation to the log4cxx source root, set a variable that contains the log4cxx root.
+set(LOG4CXX_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+
 # define the SO Version of the library. Version of the log4cxx project is subtly different than the version of the library
 # (at least in the past it was project vesion: 0.a.b.c and the library so vesion a.b.c.0
 set(LIBLOG4CXX_LIB_VERSION ${log4cxx_VERSION_MINOR}.${log4cxx_VERSION_PATCH}.${log4cxx_VERSION_TWEAK})
--- a/src/main/cpp/CMakeLists.txt
+++ b/src/main/cpp/CMakeLists.txt
@@ -197,7 +197,7 @@ if(LOG4CXX_ABI_CHECK)
 	COMMENT "Dumping ABI symbols")
 
     add_custom_target(compare-abi ALL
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -3,6 +3,11 @@ include(src/cmake/projectVersionDetails.cmake)
 project(log4cxx VERSION ${log4cxx_VER} LANGUAGES CXX)
 include(CTest)
 
+# If you are including log4cxx from a higher-level CMake file(perhaps as a submodule?)
+# CMAKE_SOURCE_DIR will refer to the first CMakeLists.txt.  Since we need some files
+# in relation to the log4cxx source root, set a variable that contains the log4cxx root.
+set(LOG4CXX_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
```
ams-tschoening, line 9: The current implementation doesn't make that variable available in the GUI of Visual Studio:

Bild_2021-04-17_111115

Changing that to the following line instead does and that approach is used for some other variables like e.g. `LOG4CXX_TEST_PROGRAM_PATH` as well.

    set(LOG4CXX_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}" CACHE PATH "Source root of log4cxx")
    set( LOG4CXX_TEST_PROGRAM_PATH "" CACHE PATH "Extra path for test programs" )

Bild_2021-04-17_111654

As this option is pretty important, shouldn't it be made as prominent as possible for people wanting to customize it?

ams-tschoening (changes requested):

CMakeLists.txt
swebb2066: I do not believe any other value of LOG4CXX_SOURCE_DIR is going to work. 

Hence I suggest not making it available in the CMake GUI

rm5248: My understanding from Steven's email was that if you have a situation like the following:

```
/MyProject
/MyProject/CMakeLists.txt
/MyProject/log4cxx
/MyProject/log4cxx/CMakeLists.txt
```

where log4cxx is some kind of submodule to MyProject that you add with some CMake directive such as `add_directory`, then CMAKE_SOURCE_DIR will be set to /MyProject, which is an issue for how log4cxx finds certain files.  For example,  the ABI dump file(relative to the log4cxx root) is src/main/abi-symbols/abi.dump.  In this case, I think it would try to look in MyProject/src/main/abi-symbols/abi.dump, instead of the intended location of MyProject/log4cxx/src/main/abi-symbols/abi.dump.

Because of that, I don't believe that it needs to be configurable.  In the normal case of building log4cxx by itself, CMAKE_SOURCE_DIR will equal CMAKE_CURRENT_SOURCE_DIR at the top-level CMakeLists file.

I'm not positive about that though, so Steven's insight would be useful.

swebb2066: I agree with Robert - it should not be configurable.

ams-tschoening: I've simply didn't read correctly, `CMAKE_CURRENT_SOURCE_DIR` vs. `CMAKE_SOURCE_DIR`, I've read the latter instead of the former. What you say makes sense of course, no need to change things.

ams-tschoening (approved):

PR #64, opened 2021-04-18
LOGCXX-523 Call error handler if rollover fails

rm5248:
If the rollover fails, call the specified error handler for
the given appender.  This means that you can now replace appenders
with a new appender.

The entire API at this point is not very robust at the moment.  A
better error handling mechanism needs to be made to properly
handle the errors so that the error handler can have enough
data to do its job.

### DIFF
--- a/src/main/cpp/rollingfileappender.cpp
+++ b/src/main/cpp/rollingfileappender.cpp
@@ -308,7 +308,9 @@ bool RollingFileAppenderSkeleton::rolloverInternal(Pool& p)
 								catch (std::exception& ex)
 								{
 									LogLog::warn(LOG4CXX_STR("Exception on rollover"));
-									errorHandler->error(ex.what(), ex, 0);
+									LogString exmsg;
+									log4cxx::helpers::Transcoder::decode(ex.what(), exmsg);
+									errorHandler->error(exmsg, ex, 0);
 								}
 							}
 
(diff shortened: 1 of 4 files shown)


rm5248:
This superceeds PR #62 

See the JIRA issue for a few caveats: https://issues.apache.org/jira/projects/LOGCXX/issues/LOGCXX-523

src/main/cpp/rollingfileappender.cpp
```diff
@@ -308,7 +308,9 @@ bool RollingFileAppenderSkeleton::rolloverInternal(Pool& p)
 								catch (std::exception& ex)
 								{
 									LogLog::warn(LOG4CXX_STR("Exception on rollover"));
-									errorHandler->error(ex.what(), ex, 0);
+									LogString exmsg;
+									log4cxx::helpers::Transcoder::decode(ex.what(), exmsg);
+									errorHandler->error(exmsg, ex, 0);
```
ams-tschoening, line 314: Another good example for maybe letting `decode` return a value in future, like [discussed](https://github.com/apache/logging-log4cxx/pull/60#issuecomment-814452988) already. Besides that, using the DECODE-macro might be easier here:

    LOG4CXX_DECODE_CHAR(exmsg, ex.what());

ams-tschoening (approved):

PR #67, opened 2021-08-16
Fix example in usage.md missing semicolon

Hex052:


ams-tschoening:
@rm5248 How about the individual logging statements? Don't they need a semicolon as well? All of those macros have been changed with 0.10.X already:

> Regarding the behavior of the LOG4CXX_INFO macro and friends: sorry, but I'm afraid it's impossible to make these macro's behave as ordinary statements without breaking some existing code. Any breakage will be clear however (compiler error) and easy to fix (**by adding a semicolon**). On the other hand, with the current macro's it is possible to create hard to find dangling else problems.

https://issues.apache.org/jira/projects/LOGCXX/issues/LOGCXX-319

This PR might be a good chance to change all of these lines as well:

> LOG4CXX_INFO(logger, "Entering application.")
vs.
> LOG4CXX_INFO(logger, "Entering application.");

rm5248 (approved):
I added a few semicolons to the needed lines and pushed those changes.

Some of these examples should be broken out into their own separate programs at some point; I think some of them won't compile as-is, and having them as full standalone examples makes it easier to modify and learn.

ams-tschoening (approved):

PR #70, opened 2021-09-08
Added ability to configure thread functions through configuration file

rm5248:
As requested in LOGCXX-431 by @coldtobi, the new thread configuration blocks signals on POSIX by default.

A new option for the configuration files has also been added, so that you may configure the behavior without recompiling.

### DIFF
--- a/src/main/cpp/domconfigurator.cpp
+++ b/src/main/cpp/domconfigurator.cpp
@@ -44,6 +44,7 @@
 #include <log4cxx/helpers/charsetdecoder.h>
 #include <log4cxx/net/smtpappender.h>
 #include <log4cxx/helpers/messagebuffer.h>
+#include <log4cxx/helpers/threadutility.h>
 
 #define LOG4CXX 1
 #include <log4cxx/helpers/aprinitializer.h>
@@ -113,6 +114,7 @@ IMPLEMENT_LOG4CXX_OBJECT(DOMConfigurator)
 #define STRINGSTREAM_ATTR "stringstream"
 #define CONFIG_DEBUG_ATTR "configDebug"
 #define INTERNAL_DEBUG_ATTR "debug"
+#define THREAD_TYPE_ATTR "threadConfiguration"
 
 DOMConfigurator::DOMConfigurator()
 	: props(), repository()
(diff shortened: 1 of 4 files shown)


src/main/cpp/domconfigurator.cpp
```diff
@@ -1035,6 +1037,19 @@ void DOMConfigurator::parse(
 		MessageBufferUseStaticStream();
 	}
 
+	LogString threadSignalValue = subst(getAttribute(utf8Decoder, element, THREAD_TYPE_ATTR));
+	if( !threadSignalValue.empty() && threadSignalValue != NuLL ){
```
ams-tschoening, line 1041: While it's not strictly related to your PR, `NuLL` reads a bit surprising to me. How about changig that to something like `[STR_]NULL[_STR]` instead?

src/main/cpp/domconfigurator.cpp
```diff
@@ -113,6 +114,7 @@ IMPLEMENT_LOG4CXX_OBJECT(DOMConfigurator)
 #define STRINGSTREAM_ATTR "stringstream"
 #define CONFIG_DEBUG_ATTR "configDebug"
 #define INTERNAL_DEBUG_ATTR "debug"
+#define THREAD_TYPE_ATTR "threadConfiguration"
```
ams-tschoening, line 117: Opinion: `THREAD_CONFIG_ATTR` seems a more reasonable name to me. `TYPE` reads like different types of threads, daemon threads vs. foreground threads vs. ..., while it's actually about their config.

src/site/markdown/threading.md
```diff
@@ -88,6 +93,27 @@ call, as there is an internal mutex that is locked and unlocked in order to ensu
...
+To tell Log4cxx what to do by default when starting a new thread, the enum
+[ThreadConfigurationType](@ref log4cxx.helpers.ThreadConfigurationType) may be
+used to configure the library appropriately.  By default, all signals on POSIX
+systems will be blocked to ensure that other threads do not get signals.
+
+To change this default, a simple change to your configuration files may be done.
+
+Example with XML configuration:
+```
+<log4j:configuration threadConfiguration="BlockSignalsOnly">
```
ams-tschoening, line 107: Opinion: As `BlockSignalsOnly` is the default already and you talk about changing that, how about using `NameThreadOnly` or some other option as example instead? That's what I expected after reading your text. As you have two examples, you could as well use two different keywords.

ams-tschoening (changes requested):
Looks good to me, still have some suggestions.

src/main/cpp/domconfigurator.cpp
rm5248: yeah that's probably a good idea - it looked odd to me too at first, and then I realized it is to avoid conflicts with the NULL macro.

src/site/markdown/threading.md
rm5248: Changing it to be the not default makes sense to me.  The reasoning behind having two examples with the same setting is that the only difference is the config file format, you're not changing two things at a time.

ams-tschoening (approved):

PR #72, opened 2021-09-18
add color to log4cxx

craterone:
Wrapping log with color and adding color pattern.
example of `ConversionPattern`:
```
log4j.appender.A1.layout.ConversionPattern=%Y[ %p ] |%d{yyyy-MM-dd hh:mm:ss}| %c -%y %m%n
```
It would set  `[ %p ] |%d{yyyy-MM-dd hh:mm:ss}| %c -` with color, and use `%Y` as start and `%y` as end.

the color with different levels
```c++
  case log4cxx::Level::FATAL_INT:
      toAppendTo.append("\x1B[35m"); //magenta
      break;
    case log4cxx::Level::ERROR_INT:
      toAppendTo.append("\x1B[91m"); //red
      break;
    case log4cxx::Level::WARN_INT:
      toAppendTo.append("\x1B[33m"); //yellow
      break;
    case log4cxx::Level::INFO_INT:
      toAppendTo.append("\x1B[32m"); //green
      break;
    case log4cxx::Level::DEBUG_INT:
      toAppendTo.append("\x1B[36m"); //cyan
      break;
    case log4cxx::Level::TRACE_INT:
      toAppendTo.append("\x1B[34m"); //blue
```


ams-tschoening:
It would be great to not only put a bunch of code here, but as well describe your use-case, the chosen implementation, limitations, how things are used in the end etc.

rm5248:
Looks good to me, but I agree with Thorsten that at least some documentation is needed(currently, the documentation for the conversion characters is in the documentation for the `PatternLayout` class).  There are a few errors on Windows(you missed a few LOG4CXX_STR macros on string constants).

Also, a unit test would be good to verify that it works correctly.  There are already some tests in patternlayouttest.cpp that you can add to.  `test1` in that file is probably the easiest to think about - there needs to be a (known-good) file in the witness/ folder that is compared to the given output.

craterone:
> Looks good to me, but I agree with Thorsten that at least some documentation is needed(currently, the documentation for the conversion characters is in the documentation for the `PatternLayout` class). 

Ok, I will add some documentation
>There are a few errors on Windows(you missed a few LOG4CXX_STR macros on string constants).

Sorry, I don't figure it out. Could you give me a hint?

> Also, a unit test would be good to verify that it works correctly. There are already some tests in patternlayouttest.cpp that you can add to. `test1` in that file is probably the easiest to think about - there needs to be a (known-good) file in the witness/ folder that is compared to the given output.

Got it, I'll add a new test.

rm5248:
> > There are a few errors on Windows(you missed a few LOG4CXX_STR macros on string constants).
> 
> Sorry, I don't figure it out. Could you give me a hint?
> 

You can see the build output here:
https://github.com/apache/logging-log4cxx/pull/72/checks?check_run_id=3641130796#step:7:120

An example:
https://github.com/apache/logging-log4cxx/pull/72/files#diff-01f067a98edee49220add01cefc76be9ffd6d17579273bc6dced2fcd0b754c25R53

Both format functions use raw string literals( `toAppendTo.append("STRING")`), that needs to be wrapped in LOG4CXX_STR to compile on Windows.

rm5248:
The code looks good to me at this point - if you could just add some documentation to `patternlayout.h` documenting the new conversion characters, I'll go ahead and merge it.

craterone:
I'll do it as soon as possible, a little busy this week.

PR #73, opened 2021-10-19
Throughput tests

swebb2066:
This PR provides a tool for measuring the cost of changes to the Logger and Hierarchy classes.

I have measured the cost of the std::mutex lock in Hierarchy::isDisabled as approx 20% of the overhead when the logger is disabled.

I will do a separate PR for  Hierarchy::isDisabled if required.


rm5248:
Is the objective to have a unit test, or a benchmarking test?  Those are two somewhat differing goals in my mind - a unit test makes sure that things are working, while a benchmarking test shows how fast under ideal circumstances.  Doing this as a unit test doesn't make much sense to me, since it really doesn't test much in terms of a regression.

FWIW here's my repo with my benchmarking tools; at some point I do want to merge it in to do continuous benchmarking tests: https://github.com/rm5248/log4cxx-bench

rm5248:
I've added my performance tests in a separate PR(#87).  I think it is equivalent to this PR and covers more cases.  If you are able to take a look at it and provide feedback, I'd appreciate it.  I did add a test to check multithreaded performance when logging is disabled(based off of this PR), so it has been useful.

swebb2066:
This has been superceded by #87

PR #75, opened 2021-11-25
Add "f" as the short filename

craterone:
As "F" in pattern layout would be a full absolute path, but I would like to log a "short" filename. I use "f" to indicate the short one.
 
For example:

using "F"
```
log4j.appender.testAppender.layout.ConversionPattern=%-5p %F(%L): %m%n
log--->
DEBUG /home/user/xxxxx/patternlayouttest.cpp(X): Message 0
```

using "f"
```
log4j.appender.testAppender.layout.ConversionPattern=%-5p %f(%L): %m%n
log--->
DEBUG patternlayouttest.cpp(X): Message 0
```


src/main/cpp/locationinfo.cpp
```diff
@@ -102,6 +120,14 @@ const char* LocationInfo::getFileName() const
 	return fileName;
 }
 
+/**
+ *   Return the short file name of the caller.
+ *   @returns file name, may be null.
```
ams-tschoening, line 131: `may be null` seems wrong to me when returning a non-pointer always. You mean `empty` instead?

src/main/cpp/locationinfo.cpp
```diff
@@ -102,6 +120,14 @@ const char* LocationInfo::getFileName() const
...
+ */
+const std::string LocationInfo::getShortFileName() const{
```
ams-tschoening, line 133: Wouldn't `const std::string&` make more sense here? This way one doesn't rely on compilers optimizing copies away and returning a reference seems always safe to me. The other getter even returns `const char*`, though, I prefer references if objects are available anyway already.

ams-tschoening (changes requested):

ams-tschoening:
Thanks for adding this and having that option seems fine to me. Though: I don't like the current wording about file name vs. short file name vs. actually being some file path. Files even sort badly right now.

Do we have any chance to rename things to what they really are most likely? A file path, a file name, a method name etc. This might result in `FilePathLocation...` and `FileNameLocation...`. More descriptive pattern characters might even be introduced like `fn` and `fp` for file name and file path.

The only problem I see right now is with `getFileName` being already available and possibly returning a path instead of a file name only in the past. This would be a breaking change for people relying on a path being available. Though, with some different pair of getters like `getFname` and `getFpath` this method might be kept and deprecated or alike.

Opinions? Does anyone think it's wort it?

src/main/cpp/locationinfo.cpp
rm5248: This should probably be a `const char*` for consistency.  There's a probably slightly more efficient way to do this as well without relying on `std::string`

src/main/cpp/locationinfo.cpp
```diff
@@ -35,6 +41,21 @@ const LocationInfo& LocationInfo::getLocationUnavailable()
...
+ * @return shortFileName
+ */
+const std::string filterFileName(const char* const fileName){
+	std::string fullFileName(fileName);
+	std::size_t found = fullFileName.rfind(SHORT_FILENAME_SPLIT_CHAR);
+	if (found != std::string::npos) {
+		return fullFileName.substr(found + 1);
+	} else {
+		return std::string(fileName);
+	}
```
rm5248, line 56: As per Thorsten's comments, this should probably do more of a `const char*` thing.  We can be somewhat clever here - since we know that the actual filename is going to be at the end of the filename, we just need to walk the string until we get to the end.  Something like this should work:

```
const char* begin = fileName;
size_t pos = 0;
while( fileName[pos] != 0 ){
  if( fileName[pos] == SHORT_FILENAME_SPLIT_CHAR &&
      fileName[pos+1] != 0 ){
      begin = fileName + pos + 1;
  }
  pos++;
}
return begin;
```

This also means that we don't _have_ to add a new member variable(this could be the implementation of `getShortFileName`), but adding a new member variable is probably good(I think with C++17 we can actually make this `constexpr` to evaluate at compile-time, but we would need to add a check for that).

rm5248:
> Thanks for adding this and having that option seems fine to me. Though: I don't like the current wording about file name vs. short file name vs. actually being some file path. Files even sort badly right now.
> 
> Do we have any chance to rename things to what they really are most likely? A file path, a file name, a method name etc. This might result in `FilePathLocation...` and `FileNameLocation...`. More descriptive pattern characters might even be introduced like `fn` and `fp` for file name and file path.
> 
> The only problem I see right now is with `getFileName` being already available and possibly returning a path instead of a file name only in the past. This would be a breaking change for people relying on a path being available. Though, with some different pair of getters like `getFname` and `getFpath` this method might be kept and deprecated or alike.
> 
> Opinions, does anyone think it's worth it?

If we want to rename things, doing it as part of a general cleanup would be the best time to do it.  There are a number of changes that I've made in the LOGCXX-516 branch that do break things in order to make things ABI-stable - my hope is to create a bigger release that cleans up a lot of the code, so that we have one bigger release that should be stable moving forward.

src/main/cpp/locationinfo.cpp
ams-tschoening: If C++17 is an option, [basic_string_view::rfind](https://en.cppreference.com/w/cpp/string/basic_string_view/rfind) could be used. Avoiding `std::string::rfind` doesn't seem that important to me, but if so, the look should be backwards. Something like the following:

```cpp
size_t len = strlen(fileName);

for (size_t pos = std::max(len - 1, 0); pos >= 0; --pos)
{
  if (fileName[pos] == SHORT_FILENAME_SPLIT_CHAR)
  {
    return fileName[pos + 1];
  }

  if (pos == 0)
  {
    return fileName;
  }
}
```

ams-tschoening:
> If we want to rename things, doing it as part of a general cleanup would be the best time to do it.

So create an issue instead for the renaming, linking to this discussion?

craterone:
> Thanks for adding this and having that option seems fine to me. Though: I don't like the current wording about file name vs. short file name vs. actually being some file path. Files even sort badly right now.

I don't like it either, but it's really hard to find a better name :(.

> So create an issue instead for the renaming, linking to this discussion?

I think it's better to create a new issue.

src/main/cpp/locationinfo.cpp
Sadboipoor: N

src/main/cpp/locationinfo.cpp
Sadboipoor: Sad

Sadboipoor (approved):
H

rm5248:
Created PR #95 off of this one and merged in; closing this PR.  Thank you for the patch.

PR #81, opened 2021-12-11
Don't convert CRLF, git

coldtobi:
By default (at least on Linux) git does convert CRLF automatically, also in the
above mentioned file. However, that CRLF is a feature in that file....

Jira: LOGCXX-540


ams-tschoening:
Seems that some editors like my Sublime Text 3 don't even show something like mixed line endings, but it claims the file to be `Unix` only. The line of interest is the following:

```
propertiestestcase.crlf1=continued\
 value
```

image

ams-tschoening (changes requested):
I would like to suggest a slightly different code style, otherwise am fine with the change.

```gitattributes
*     text eol=lf
*.png binary
*.jpg binary

# This file has a CRLF that are important and must not be changed by GIT.
src/test/resources/input/propertiestestcase.properties -text

# This is a windows-specific file.
src/main/resources/log4cxx.rc eol=crlf
```

rm5248:
I was actually just about to go and fix this by creating a new file that was in Windows(CRLF) line endings, since that particular property is only used in [one test.](https://github.com/apache/logging-log4cxx/blob/d08a933041c87c7a68908334df5b0cceadd3a7f5/src/test/cpp/helpers/propertiestestcase.cpp#L158)  Is there some reason we want to have a file with mixed line endings?  Keeping it separate seems like the best idea to me.

ams-tschoening:
If it doesn't break anything else, agreed, your approach would be better. Especially if that use-case is already covered in `.gitattributes`.

coldtobi:
I'm not sure about what changes requested: 
- The suggestion from @ams-tschoening for formatting the gitattributes file
- rm5248's comment

I thought that the purpose of the testcase is to test both possiblities of line continuations (on Win and *nix)

rm5248:
> I thought that the purpose of the testcase is to test both possiblities of line continuations (on Win and *nix), as the properties are named accordingly (the extra crlf in the properties name)

There's one line in the properties file per test-case.  So it does test `\r\n` on *nix.  There's one line in the properties file per test-case defined.

My proposal is to simply create a new file(propertiestestcase-crlf.properties or something like that) that only has one line in it(apart from the standard Apache header):
```
propertiestestcase.crlf1=continued\^M
 value
```
That way, we can set the .gitattributes file to set propertiestestcase-crlf.properties to be `eol=crlf`:
```
src/main/resources/propertiestestcase-crlf.properties eol=crlf
```

ams-tschoening:
> I thought that the purpose of the testcase is to test both possiblities of line continuations (on Win and *nix), as the properties
> are named accordingly (the extra crlf in the properties name)

Having a second thought, that's how I understand it as well now:

```
propertiestestcase.tab7=continued\
	value
propertiestestcase.crlf1=continued\
 value
```

The first of those properties is not only TAB-focused, but LF as well, otherwise the continued values for those both wouldn't make too much sense.

ams-tschoening:
> There's one line in the properties file per test-case. So it does test \r\n on *nix.[...]

Which means currently the properties parser is tested with mixed line endings as well, which wouldn't be the case anymore if two different files with their unique line endings would be used?

rm5248:
> > There's one line in the properties file per test-case. So it does test \r\n on *nix.[...]
> 
> Which means currently the properties parser is tested with mixed line endings as well, which wouldn't be the case anymore if two different files with their unique line endings would be used?

That looks like more of an unintentional side-effect of the test, rather than something that is checked explicitly.  I think this needs to be split into the following parts, so that we test everything explicitly:

- The current test(minus the current crlf property)
- `\n` on all platforms(propertiestsetcase-lf.properties)
- `\r\n` on all platforms(propertiestestcase-crlf.properties)
- Mixed line endings(propertiestestcase-mixed.properties)

rm5248:
I've made a new PR with the changes, if you are able to review that would be helpful: https://github.com/apache/logging-log4cxx/pull/85

coldtobi:
> I've made a new PR with the changes, if you are able to review that would be helpful: #85

#85 works for me; I guess then lets close this PR, as #85 succeeds it.

PR #83, opened 2021-12-13
Fix constructions of sed filter in corner cases

coldtobi:
Some tests use sed to generate predictable outputs and employ sed files for
that.  In some circumstances, the generated sed file syntax is incorrect, if
the subsitions data contains the sed expresssion seperator. The patch escapes
the data if needed.  The bug manifested in (local) autopkgtest runs, when the
temporary directory contained the separator ("Q").

https://issues.apache.org/jira/browse/LOGCXX-543

### DIFF
--- a/src/test/cpp/util/transformer.cpp
+++ b/src/test/cpp/util/transformer.cpp
@@ -116,7 +116,7 @@ void Transformer::createSedCommandFile(const std::string& regexName,
 
 	std::string tmp;
 
-	auto sedsaniziter = [] (const std::string in, const std::string sedseperator = "Q") {
+	auto sedsaniziter = [] (const std::string &in, const std::string &sedseperator = "Q") {
 		std::string ret = in;
 		std::string replace_to = "\\" + sedseperator;
 		size_t pos = 0;
--- a/src/test/cpp/util/transformer.cpp
+++ b/src/test/cpp/util/transformer.cpp
@@ -116,14 +116,25 @@ void Transformer::createSedCommandFile(const std::string& regexName,
 
 	std::string tmp;
 
+	auto sedsaniziter = [] (const std::string in, const std::string sedseperator = "Q") {
+		std::string ret = in;
+		std::string replace_to = "\\" + sedseperator;
+		size_t pos = 0;
+		while((pos = ret.find(sedseperator, pos)) != std::string::npos) {
+			ret.replace(pos, sedseperator.length(), replace_to);
+			pos += replace_to.length();
+		}
+		return ret;
+	};
+
 	for (log4cxx::Filter::PatternList::const_iterator iter = patterns.begin();
 		iter != patterns.end();
 		iter++)
 	{
 		tmp = "sQ";
-		tmp.append(iter->first);
+		tmp.append(sedsaniziter(iter->first));
 		tmp.append(1, 'Q');
-		tmp.append(iter->second);
+		tmp.append(sedsaniziter(iter->second));
 		tmp.append("Qg\n");
 		apr_file_puts(tmp.c_str(), regexFile);
 	}
--- a/src/test/cpp/util/transformer.cpp
+++ b/src/test/cpp/util/transformer.cpp
@@ -116,15 +116,15 @@ void Transformer::createSedCommandFile(const std::string& regexName,
 
 	std::string tmp;
 
-	auto sedSanitizer = [] (const std::string& in, const std::string& sedSeperator = "Q")
+	auto sedSanitizer = [] (const std::string& in, const std::string& sedSeparator = "Q")
 	{
 		std::string ret = in;
-		std::string replaceTo = "\\" + sedSeperator;
+		std::string replaceTo = "\\" + sedSeparator;
 		size_t pos = 0;
 
-		while((pos = ret.find(sedSeperator, pos)) != std::string::npos)
+		while((pos = ret.find(sedSeparator, pos)) != std::string::npos)
 		{
-			ret.replace(pos, sedSeperator.length(), replaceTo);
+			ret.replace(pos, sedSeparator.length(), replaceTo);
 			pos += replaceTo.length();
 		}
 
--- a/src/test/cpp/util/transformer.cpp
+++ b/src/test/cpp/util/transformer.cpp
@@ -116,14 +116,18 @@ void Transformer::createSedCommandFile(const std::string& regexName,
 
 	std::string tmp;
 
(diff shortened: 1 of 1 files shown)


src/test/cpp/util/transformer.cpp
```diff
@@ -116,14 +116,25 @@ void Transformer::createSedCommandFile(const std::string& regexName,
 
 	std::string tmp;
 
+	auto sedsaniziter = [] (const std::string in, const std::string sedseperator = "Q") {
```
ams-tschoening, line 119: Shouldn't the arguments better be `const std::string&`? Especially as you are copying `in` again.

coldtobi: you're right, fixing ...

src/test/cpp/util/transformer.cpp
```diff
@@ -116,14 +116,25 @@ void Transformer::createSedCommandFile(const std::string& regexName,
 
 	std::string tmp;
 
+	auto sedsaniziter = [] (const std::string &in, const std::string &sedseperator = "Q") {
```
ams-tschoening, line 119: `sedsaniziter` contains a spelling error most likely, prefer `sedSanitizer`. `sedseperator` should follow that naming with `sedSeperator` as well. We mostly follow Java naming conventions.

Please change the references to be `const std::string& ...`, because most other code does so already.

src/test/cpp/util/transformer.cpp
```diff
@@ -116,14 +116,25 @@ void Transformer::createSedCommandFile(const std::string& regexName,
...
+		std::string ret = in;
+		std::string replace_to = "\\" + sedseperator;
+		size_t pos = 0;
+		while((pos = ret.find(sedseperator, pos)) != std::string::npos) {
```
ams-tschoening, line 123: Please add a newline before and after the `while`, make things better readable.

ams-tschoening (changes requested):

src/test/cpp/util/transformer.cpp
coldtobi: should be fixed now by 38729af.

src/test/cpp/util/transformer.cpp
coldtobi: should be fixed now by 38729af.

src/test/cpp/util/transformer.cpp
```diff
@@ -116,14 +116,18 @@ void Transformer::createSedCommandFile(const std::string& regexName,
 
 	std::string tmp;
 
-	auto sedsaniziter = [] (const std::string &in, const std::string &sedseperator = "Q") {
+	auto sedSanitizer = [] (const std::string& in, const std::string& sedSeperator = "Q")
```
ams-tschoening, line 119: Sorry, missed that as well: Needs to be `sedSeparator` using an `a`.

ams-tschoening (changes requested):

ams-tschoening (approved):
Things look good and tests still work on my Windows.

src/test/cpp/util/transformer.cpp
```diff
@@ -116,14 +116,29 @@ void Transformer::createSedCommandFile(const std::string& regexName,
...
+		}
+
+		return ret;
+	};
+
 	for (log4cxx::Filter::PatternList::const_iterator iter = patterns.begin();
 		iter != patterns.end();
 		iter++)
 	{
 		tmp = "sQ";
```
ams-tschoening, line 138: For my own thoughts: This results in something like `sQ...Q...Qg`, which is pretty much the same like `s/.../.../g` or `s!...!...!g` or `s#...#...#g` etc. The important point is that at least the delimiter needs to be escaped when available in the input, like was the case for this PR.

OTOH, in theory each and every character would need to be escaped because it might have special reg exp meaning? SED documents things like `&` and `\1` already. Though, might not be that important in this context, as it's only about tests with the only variable input being some paths in the file system.

Modern versions of SED support the argument `--regexp-extended`, which should support `\Q...\E` to escape arbitrary input. So something like `sQ\Q...\EQ\Q...\EQg` should work in the future as well.

https://www.boost.org/doc/libs/1_50_0/libs/regex/doc/html/boost_regex/syntax/basic_extended.html#boost_regex.syntax.basic_extended.quoting_escape

Though, a different delimiter should be chosen than to keep things readable. But on my system I have a version of SED using `UnxUtils` for Windows which doesn't support that. Instead, I would need to use SED bundled with GIT, but sadly wasn't able to get that work as expected as well. So am simply keeping things as they are right now.

> C:\Program Files (x86)\UnxUtils\usr\local\wbin

vs.

> C:\Program Files\Git\usr\bin

[ghpr_83_full_escape.diff.txt](https://github.com/apache/logging-log4cxx/files/7721553/ghpr_83_full_escape.diff.txt)

coldtobi: Yes, this PR is only focusing on the delimiter only -- to fix the test issues I saw when packaging... 

Regarding your sed: Did you try the switch `-E` instead? `--regexp-extend` is not POSIX, but `-E` would be, so maybe it understands only that switch (TIL: the feature is indeed POSIX; before, I thought this was not a portable feature)

Of course, if wanted, I can extend sedSanitizer  to escape more characters, so that it `-E` (aka `--rexep-extended` is not needed), or look into boost::regexp to avoid the invocation to sed alltogether. Though this feels for me as overkill just for the tests (having to have boost I mean)

ams-tschoening: I've tested and `--regexp-extend` is supported by my GIT-SED. Though, it seems I have missed again that regular expression are used and need to be supported, making my thoughts somewhat useless. :-) So let's simply keep things as they are now and change when needed as needed.

rm5248: > Of course, if wanted, I can extend sedSanitizer to escape more characters, so that it `-E` (aka `--rexep-extended` is not needed), (Edit: Won't work for the same reason as in the last paragraph) or look into boost::regexp to avoid the invocation to sed alltogether. Though this feels for me as overkill just for the tests (having to have boost I mean)

FWIW since we're now on C++11, we have `std::regex`, which is basically the same as `boost::regexp`.  If your compiler doesn't support C++17, boost is currently required anyway.

coldtobi: Gosh, I thought std.:regex was C++17 ... I stand corrected.

PR #84, opened 2021-12-14
Fix underspecification for logger

fulldecent:
(Please update this PR to fix all affected functions in all files. This is zero-day live Twitch stream, can't type more here.)

---

This function is specified as logging a string.

In light of recent disclosures, actually much more is happening here than interpreting only as "message the message string to log."

### DIFF
--- a/src/main/include/log4cxx/logger.h
+++ b/src/main/include/log4cxx/logger.h
@@ -174,7 +174,7 @@ class LOG4CXX_EXPORT Logger :
 		registered appenders in this logger and also higher in the
 		hierarchy depending on the value of the additivity flag.
 
-		@param msg the message string to log.
+		@param msg the message string to log (WARNING, THIS STRING MAY RESULT IN CODE BEING EXECUTED AS SHELL).
 		@param location location of source of logging request.
 		*/
 		void debug(const std::string& msg, const log4cxx::spi::LocationInfo& location) const;


rm5248:
I'm assuming that you're referring to CVE-2021-44228 in log4j2.  Log4cxx does not(and as far as I am aware has never) supported any sort of parameter lookup inside of log messages, and is immune to JNDI lookup as it does not use Java.  Is there something else that you are referring to?

ams-tschoening:
> [...]This is zero-day live Twitch stream, can't type more here.

I see how important things are for you... :-) Let's close, reads like nonsense to me. It's not even shell code that gets executed in the first place AFAIK.

fulldecent:
Coming back here with more time to type.

This issue discussed here affects:

- The Apache Log4j 2 API specification
- Every Apache Log4j 2 implementation

The distinction between these two things is discussed at: "API Separation" in https://logging.apache.org/log4j/2.x/

---

# Here is the problem:

On the marketing page https://logging.apache.org/log4j/2.x/ it is stated:

> The Log4j API supports logging Messages instead of just Strings

This is a normative statement which significantly describes methods in log4cxx and every other implementation.

That means that any API documentation and implementation of Log4j 2 (including log4cxx) is expected to support "Messages".

The documentation in the line referenced in this PR (as well as all other logging methods in this implementation, all other logging methods in all other implementations, and in the specification itself) MUST (meaning defined in RFC 2119) support Messages.

The above-referenced API documentation mentions only "messages".

What is the difference between a "message" and a "Message"? Well the lowercase version has the plain English meaning of approximately "something which is said". And the uppercase version has a meaning specified at https://logging.apache.org/log4j/2.x/manual/messages.html

# Required fix for API conformance

To be compliant with normative references in the authoritative API specifications, the above-cited function must expanded to support Messages, not just strings.

That includes updating the specification (function documentation), and possibly the implementation. If this implementation does not support Messages, then it is not fully conformant with the API specification and such deficiency should be specifically noted.

# Recommended fix for safety

In addition to the minimal change of making one of the letters uppercase, it can be very helpful to provide context and related useful documentation for what Messages means.

That could include linking to the normative page at https://logging.apache.org/log4j/2.x/manual/messages.html or possibly more context-specification (i.e. related only to this implementation) details on what messages are supported.

ams-tschoening:
> This issue discussed here affects:[...]

Log4cxx is neither API compatible with Log4j2 nor an implementation of that. AFAIK Log4cxx doesn't even claim to be so anywhere.

> This is a normative statement which significantly describes methods in log4cxx and every other implementation.

No it doesn't.

> That means that any API documentation and implementation of Log4j 2 (including log4cxx)[...]

That claim is wrong.

> [...]MUST (meaning defined in RFC 2119) support Messages.

No it doesn't.

> Required fix for API conformance

Log4cxx doesn't claim to be API compatible with Log4j2 and of course doesn't need to. No logging framework needs to be and can freely decide that on its own. Your PR doesn't make any sense:

It speaks about STRINGS in the added comment, while you are now discussing somwthing about `Messages` vs. `messages` vs. strings and are trying to convince people to implement one over the other for crude reasons. Besides that, you are claiming that shell code can be executed in your PR without proving that in any way. And even if Log4cxx would implement `Messages` like designed by Log4j2, that itself wouldn't necessarily mean that Log4cxx would be vulnerable in the same way like Log4j2 is with Log4Shell.

Closing, because this doesn't seem to lead anywhere useful. @rm5248 Feel free to correct me. :-)

fulldecent:
Thank you. Sorry for my confusion here.

Yes, I see that Log4cxx is NOT an implementation of Log4j2. They are different things. The only thing the same is the inspiration and similar names.

I have moved this issue to the correct place at https://github.com/apache/logging-log4j2/pull/630

PR #88, opened 2021-12-27
Improve throughput when using disabled logging statements

swebb2066:
Change the pointer to the Hierarchy from a std::weak_ptr to a raw pointer to avoid the lock() in Logger::isDisabled() functions.

Also disconnects the Logger from the Hierarchy when the Hierarchy is destroyed.

**throughputtests output using these changes:**
```Benchmarking library only(no writing out):
**************************************************************
Benchmarking Single threaded: 1000000 messages
**************************************************************
Log4cxx NoFormat pattern: %m%n Elapsed: 2.557 secs 391,157/sec
Log4cxx DateOnly pattern: [%d] %m%n Elapsed: 2.552 secs 391,836/sec
Log4cxx DateClassLevel pattern: [%d] [%c] [%p] %m%n Elapsed: 2.572 secs 388,729/sec
Log4cxx Logging with FMT Elapsed: 1.909 secs 523,721/sec
Log4cxx Logging static string Elapsed: 2.062 secs 485,083/sec
Log4cxx Logging static string with FMT Elapsed: 2.096 secs 476,990/sec
Log4cxx Logging disabled debug Elapsed: 0.08675 secs 11,527,275/sec
Log4cxx Logging disabled trace Elapsed: 0.08727 secs 11,459,057/sec
Log4cxx Logging enabled debug Elapsed: 2.059 secs 485,615/sec
Log4cxx Logging enabled trace Elapsed: 2.098 secs 476,611/sec
**************************************************************
Benchmarking multithreaded threaded: 1000000 messages/thread, 4 threads
**************************************************************
log4cxx: No appender could be found for logger (bench_logger).
log4cxx: Please initialize the log4cxx system properly.
Log4cxx Logging with FMT MT Elapsed: 3.734 secs 267,813/sec
Log4cxx Logging with FMT MT Elapsed: 3.739 secs 267,426/sec
Log4cxx Logging with FMT MT Elapsed: 3.759 secs 266,043/sec
Log4cxx Logging with FMT MT Elapsed: 3.762 secs 265,849/sec
**************************************************************
Benchmarking multithreaded disabled: 1000000 messages/thread, 4 threads
**************************************************************
Log4cxx Logging disabled MT Elapsed: 0.2366 secs 4,227,422/sec
Log4cxx Logging disabled MT Elapsed: 0.2459 secs 4,066,714/sec
Log4cxx Logging disabled MT Elapsed: 0.2496 secs 4,006,421/sec
[...]

PR #89, opened 2021-12-27
Prevent thoughputtests warning

swebb2066:
This PR is required to prevent 'No appender could be found for logger (bench_logger)' warning in throughputtests after applying #88


swebb2066:
Suerceded by #90

PR #91, opened 2021-12-28
Allow throughputtests to be built when LOG4CXX_LOGCHAR_IS_WCHAR is on

swebb2066:
This PR completes #87


swebb2066:
Duplicates #90

PR #92, opened 2021-12-28
Allow throughputtests to be built when LOGCHAR_IS_WCHAR is on

swebb2066:
This change adorns literal strings within LOG4CXX_STR where required.

### DIFF
--- a/src/test/cpp/throughput/throughput-main.cpp
+++ b/src/test/cpp/throughput/throughput-main.cpp
@@ -48,7 +48,7 @@ static void benchmark_function( const std::string& name, void (*fn)(int), int ho
 	LOG4CXX_INFO_FMT( console, "Log4cxx {} Elapsed: {:.4} secs {:L}/sec",
 		name,
 		delta_d,
-		int(howmany / delta_d));
+		uint64_t(howmany / delta_d));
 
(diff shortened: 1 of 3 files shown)


src/test/cpp/throughput/throughput-main.cpp
```diff
@@ -71,14 +74,14 @@ static void benchmark_conversion_pattern( std::string name,
 		delta_d,
 		int(howmany / delta_d) );
 
-	results.push_back( howmany / delta_d );
+	results.push_back( uint64_t(howmany / delta_d) );
```
ams-tschoening, line 77: If `results` use `uint64_t`, shouldn't the line above use that type as well? How about the following for those two places instead, which prevents double calculation at the same time.

```cpp
	results.push_back( uint64_t(howmany / delta_d) );
	LOG4CXX_INFO_FMT( console, "Log4cxx {} pattern: {} Elapsed: {:.4} secs {:L}/sec",
		name,
		conversion_pattern,
		delta_d,
		results.back());
```

[ghpr_92.diff.txt](https://github.com/apache/logging-log4cxx/files/7787940/ghpr_92.diff.txt)

ams-tschoening (changes requested):

PR #95, opened 2022-01-02
Add short filename to location info

rm5248:
This is an extension of PR #75.  This will calculate the short filename at compile time(assuming you have C++17) and give it to the LocationInfo class.  If you do not have C++17 and you request the short filename, it will calculate it at runtime.

Documentation on how to disable location information has also been added(with a test).


src/main/include/log4cxx/spi/location/locationinfo.h
```diff
@@ -132,9 +152,28 @@ class LOG4CXX_EXPORT LocationInfo
 #if !defined(__LOG4CXX_FUNC__)
 	#define __LOG4CXX_FUNC__ ""
 #endif
+
+#if __cpp_lib_string_view || (_MSVC_LANG >= 201703L)
+#include <string_view>
+
+#define LOG4CXX_LOCATION_CREATE ::std::string_view file_name{__FILE__};\
```
ams-tschoening, line 159: I don't get the point of this macro: Isn't this calculated at runtime as well, only differently than `getShortFileName`? Why not have two different calculations of `getShortFileName` in the class instead of a macro containing complex implementation, making debugging more difficult in case of problems? Is there some `constexpr` or `static` missing to calculate only once?

src/main/cpp/locationinfo.cpp
```diff
@@ -45,16 +45,35 @@ LocationInfo::LocationInfo( const char* const fileName1,
 	int lineNumber1 )
 	:  lineNumber( lineNumber1 ),
 	   fileName( fileName1 ),
+	   shortFileName(nullptr),
```
ams-tschoening, line 48: Shouldn't this use `LocationInfo::NA` like for the default CTOR?

src/main/cpp/locationinfo.cpp
```diff
@@ -102,6 +122,31 @@ const char* LocationInfo::getFileName() const
...
+		return shortFileName;
+	}
+
+	if(fileName == nullptr){
+		return nullptr;
+	}
+
+	const char* begin = fileName;
+	size_t pos = 0;
+	while( fileName[pos] != 0 ){
```
ams-tschoening, line 140: Opinion: Newlines before/after the loop increase readability.

ams-tschoening (changes requested):

src/main/cpp/locationinfo.cpp
```diff
@@ -45,16 +45,35 @@ LocationInfo::LocationInfo( const char* const fileName1,
...
 
+LocationInfo::LocationInfo( const char* const fileName1,
+							size_t shortFilenameOffset,
+	const char* const methodName1,
+	int lineNumber1 )
+	:  lineNumber( lineNumber1 ),
+	   fileName( fileName1 ),
+	   methodName( methodName1 )
+{
+	if( shortFilenameOffset < 0 || shortFilenameOffset > 300 ){
```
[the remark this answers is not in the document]
rm5248: My (perhaps misguided) reasoning is that because this accesses a `const char*` directly, that could let you access arbitrary memory.  I'm not sure how likely that is though.

src/main/include/log4cxx/spi/location/locationinfo.h
rm5248: Using `string_view` lets it be calculated at compile-time(assuming a high enough optimization level).  Upon some further investigation though, both GCC and MSVC can evaluate `strlen` on a static string at compile time.  GCC at least is also smart enough to evaluate `strrchr` at compile time, so I think we can do this with only C functions.

Basically the following will work in both MSVC and GCC(at least according to godbolt), with GCC managing to figure out at compile-time that 'name' is '/hithere.cpp':
```
const char* file_name = "/bar/foo/hithere.cpp";
const int file_length = strlen(file_name);
const char* name = strrchr(file_name, '/');
```

Unfortunately it seems that with MSVC using either `string_view` or `strrchr` still requires some processing at run-time in order to calculate the correct position.

src/main/cpp/locationinfo.cpp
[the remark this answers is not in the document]
ams-tschoening: If the file name calculation works, there won't be such access. If it doesn't work, it could easily happen at 199 chars as well already. OTOH, arbitrary path restrictions regarding their length, allowed characters like spaces etc. are a continuing annoyance for decades, so in my opinion we shouldn't add one unless absolutely necessary.

src/main/include/log4cxx/spi/location/locationinfo.h
ams-tschoening: So while the macro copies code in the end multiple times, with a supporting compiler the calculation might happen at compile time because of an optimized `string_view`. But that is not guaranteed and the macro might as well be executed multiple times at runtime?

That seems unpredictable and unnecessary confusing to me. Why not use a static function based on `constexpr` instead, which is C++11 and things either reliably break or work? If `string_view` is supported at compile time, shouldn't that be available with `constexpr` then as well? But that would be guaranteed to happen at compile time.

OTOH, your custom implementation might work with `constexpr` as well which is documented to have support for local [variables and loops since C++14](https://en.cppreference.com/w/cpp/language/constexpr). So I suggest something like the following:

```c++
#define LOG4CXX_LOCATION_CREATE ::log4cxx::spi::LocationInfo location(__FILE__,         \
	::log4cxx::spi::LocationInfo::calcShortFileName(__FILE__), \
	__LOG4CXX_FUNC__, \
	__LINE__)
```

The implementation of that function checks if `string_view` is available and uses that or your custom implementation. Though, with your custom implementation in backwards, `string_view` is redundant anyway. The functions documents to be calculated at compile time and the macro is less verbose as well.

rm5248: Yes, I like your idea of a separate function, that makes it much cleaner.  I was trying to avoid expanding __FILE__ multiple times, until I remembered that the compiler will only put in the same string once into the binary.  That also makes it easier to support both versions: one using `string_view` and the fallback using `strrchr`.

src/main/include/log4cxx/spi/location/locationinfo.h
```diff
@@ -45,21 +58,28 @@ class LOG4CXX_EXPORT LocationInfo
...
-
+#ifdef LOG4CXX_HAS_STRING_VIEW
+		static constexpr const char* calcShortFileName(const char* fileName){
+			std::string_view view(fileName);
+			// If the separator is not found, rfind will return -1.  Adding 1 to
+			// that will have it pointing at fileName, which is a good fallback.
+			return fileName + view.rfind(LOG4CXX_SHORT_FILENAME_SPLIT_CHAR) + 1;
+		}
+#else
+		static const char* calcShortFileName(const char* fileName){
```
ams-tschoening, line 69: Some additional docs of why `constexpr` is not possible with `strrchr`:

https://stackoverflow.com/questions/42352114/why-all-function-in-cstring-must-not-have-constexpr

src/main/include/log4cxx/spi/location/locationinfo.h
```diff
@@ -45,21 +58,28 @@ class LOG4CXX_EXPORT LocationInfo
...
+			return strrchr(fileName, LOG4CXX_SHORT_FILENAME_SPLIT_CHAR) + 1;
```
ams-tschoening, line 70: `strrchr` might return a NULL pointer in case SPLIT_CHAR is not found [according to the docs](https://www.cplusplus.com/reference/cstring/strrchr/?kw=strrchr). Shouldn't that be handled? You explicitly mention that case for `string_view`. `NULL + 1` seems wrong.

ams-tschoening (changes requested):

src/main/include/log4cxx/spi/location/locationinfo.h
rm5248: whoops you're right.  If it's found we need to add 1(like the string_view version), else we should return the `fileName` as a sane default value.

ams-tschoening (approved):
Tested on my VS as well and all tests passed.

PR #96, opened 2022-01-03
Reduce indirection in Logger and Hiearchy implementations

swebb2066:
Also remove temporary LOGCXX-322 patch that causes compilation error when APR_HAS_THREADS is set to 0

### DIFF
--- a/src/main/cpp/hierarchy.cpp
+++ b/src/main/cpp/hierarchy.cpp
@@ -53,8 +53,6 @@ struct Hierarchy::HierarchyPrivate
 {
 	HierarchyPrivate()
 	{
-		loggers = std::make_unique<LoggerMap>();
-		provisionNodes = std::make_unique<ProvisionNodeMap>();
 		root = std::make_shared<RootLogger>(pool, Level::getDebug());
 		defaultFactory = std::make_shared<DefaultLoggerFactory>();
 		emittedNoAppenderWarning = false;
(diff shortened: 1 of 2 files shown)


rm5248:
Looks good to me.

The ABI failure is not due to this change, but due to some of the other changes that you've made(e.g. returning a reference instead of the `shared_ptr`).  At some point it would be interesting to see if there is a significant speed difference returning a reference vs. the `shared_ptr` - I would be concerned that returning a reference to a `shared_ptr` (and it seems that modern compilers [may not have an issue with this anyway](https://stackoverflow.com/questions/10643563/how-to-return-smart-pointers-shared-ptr-by-reference-or-by-value)).

swebb2066:
To copy a shared_ptr the reference count must be incremented, for which a lock must be acquired.

By returning a const reference, the caller can decide whether to copy the shared_ptr. The compiler will complain when the const reference is passed to a function requiring a non-const reference.

ams-tschoening:
> To copy a shared_ptr the reference count must be incremented, for which a lock must be acquired.

The link Robert provided tells the exact opposite:

> The cost concern is nowadays moot thanks to return value optimization (RVO), so you won't incur a increment-increment-decrement sequence or something like that in modern compilers. So the best way to return a shared_ptr is to simply return by value:

rm5248:
I think it may actually depend on the context.  If you're returning a new `shared_ptr` the reference count would not have to be incremented.  If you're returning a `shared_ptr` to some sort of member variable it would need to be incremented and thus locked.

swebb2066:
To prevent unnecessary locks in #90 it was necessary to change the return type of getEffectiveLevel() to a const reference. So RVO probably does not happen for virtual functions.

It surprised me - I only found it by using perf

swebb2066:
It occurred to me that RVO is only relevent for functions that create the returned object. getEffectiveLevel() returns the LevelPtr held by theLogger.

PR #99, opened 2022-01-08
[LOG4CXX-548] Reproducible Doxygen doc generation

coldtobi:
The way the doxygen documenation is generated with CMake, it is unfortunatly
not reproducible (as in https://reproducible-builds.org/).

This is a regression as log4cxx has been previously reproducible (at least on
Debian –
https://tests.reproducible-builds.org/debian/rb-pkg/experimental/amd64/log4cxx.html)

The problem is that doxygen's configure has several absolute paths, which it
then bakes into the generated documentation, for example in this snippt, the
bold part dependent on the build path:

<div class="contents">
<div class="textblock"><p><a class="anchor" id="md__build_log4cxx_0_12_1_src_site_markdown_index"></a></p>

### DIFF
--- a/src/site/CMakeLists.txt
+++ b/src/site/CMakeLists.txt
@@ -1,15 +1,18 @@
 option(BUILD_SITE "Build log4cxx website" OFF)
 if(BUILD_SITE)
-    find_package( Doxygen REQUIRED dot )
-    configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/doxy/Doxyfile.in
-	            ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile )
+	file(RELATIVE_PATH LOG4CXX_REL_BIN_TO_SRC_DIR "${LOG4CXX_SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}")
 
-    configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/markdown/download.md.in
-	            ${CMAKE_CURRENT_BINARY_DIR}/markdown/download.md )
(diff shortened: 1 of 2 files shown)


rm5248:
The patch as given requires CMake 3.20+, but doesn't update the minimum required version.  I'd like to not have to bump the CMake version if possible.

I changed the `cmake_path` directive to `file(RELATIVE_PATH LOG4CXX_REL_BIN_TO_SRC_DIR "${LOG4CXX_SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}")` on my side which seems to work fine, would you be able to validate that?  I don't see where the particular example that you showed is coming in from, that only seems to be a problem with the MapFilter page.

coldtobi:
> The patch as given requires CMake 3.20+, but doesn't update the minimum required version. I'd like to not have to bump the CMake version if possible.

Ack.

> I changed the `cmake_path` directive to `file(RELATIVE_PATH LOG4CXX_REL_BIN_TO_SRC_DIR "${LOG4CXX_SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}")` on my side which seems to work fine, would you be able to validate that? 

LGTM, it build reproducible as well on my machine. I'll update my PR accordingly.

> I don't see where the particular example that you showed is coming in from, that only seems to be a problem with the MapFilter page.

It seems everything in site/markdown triggering the differences... Here's a (local created) complete diff: https://pastebin.com/6BLaEa0K. (So it is quite possible that only the INPUT configuration of Doxyfile.in would need to be relative, but I believe making the other relative won't hurt.)

PR #102, opened 2022-01-15
Create intermediate directories for both fixedwindowrollingpolicy and…

rm5248:
… timebasedrollingpolicy

### DIFF
--- a/src/main/cpp/rollingpolicybase.cpp
+++ b/src/main/cpp/rollingpolicybase.cpp
@@ -28,6 +28,7 @@
 #include <log4cxx/pattern/patternparser.h>
 #include <log4cxx/pattern/integerpatternconverter.h>
 #include <log4cxx/pattern/datepatternconverter.h>
+#include <log4cxx/helpers/optionconverter.h>
 
 using namespace log4cxx;
 using namespace log4cxx::rolling;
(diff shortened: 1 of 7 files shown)


src/main/cpp/rollingpolicybase.cpp
```diff
@@ -38,6 +39,7 @@ IMPLEMENT_LOG4CXX_OBJECT(RollingPolicyBase)
 
 RollingPolicyBase::RollingPolicyBase()
 {
+	createIntermediateDirectories = false;
```
ams-tschoening, line 42: While I understand using `false` for backwards compatibility, isn't the whole point of the [issue](https://issues.apache.org/jira/browse/LOGCXX-525) and PR to fix unexpected behavior/limitations? Other implementations of [appenders](https://logback.qos.ch/manual/appenders.html#FileAppender) for .e.g Logback seem to create those dirs by default as well:

> If the parent directory of the file does not exist, FileAppender will automatically create it, including any necessary but nonexistent parent directories.

rm5248: I had originally had it as `true`, but there's a [test that broke](https://github.com/apache/logging-log4cxx/blob/8988e3e7ced71000e2f1e5c9bd1e282cb766419a/src/test/cpp/rolling/manualrollingtest.cpp#L223) when I did that.  I'm not sure exactly _why_ this test checks for that, but it does.

The [log4j2 appender](https://logging.apache.org/log4j/2.x/manual/appenders.html#FileAppender) does create directories as well.

I think it's perfectly reasonable to change the default to 'true' if you agree.

ams-tschoening: @stephen-webb @swebb2066 What's your opinion on this?

PR #103, opened 2022-01-17
Reference expat xml library consistenty

swebb2066:
Currently, the examples use ${XMLLIB_LIBRARIES} while tests and the log4cxx library use EXPAT::EXPAT.

This PR aims to localise in the top level CMakeLists.txt the expat library reference. 

Note: find(EXPAT) will cornnect to the script shipped with the CMake modules while Find(expat) will connect to the scripts shipped by Expat.


rm5248:
Would it make more sense to merge this into master instead of next_stable?

Otherwise it looks good to me.

swebb2066:
It occurred to me I should hande bot types of find result.

If this is OK I will cherry-pick to a master destined PR.

rm5248:
looks good to me.

I prefer to merge into master and then merge master into next_stable manually once I have a few changes(so I don't have to do the same fix twice), and I think it keeps the history clean at that point, since the commits are in both branches.  If you want to manually cherry-pick it that works too(no need to do a separate PR).

PR #106, opened 2022-01-25
Sonarcloud build

rm5248:
Scan the library using sonarcloud

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -18,6 +18,8 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types: [opened, synchronize, reopened]
 
 jobs:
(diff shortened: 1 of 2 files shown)


ams-tschoening:
Just out of interest: What exactly does Sonarcloud scan for in this context?

rm5248:
Well it _should_ be scanning for whatever vulnerabilities that Sonarqube scans for via static code analysis, although it says there are no problems which is very suspicious to me.  No code seems to show up on the sonarcloud UI.

rm5248:
Merging to master does produce a number of(expected) bugs: https://sonarcloud.io/summary/overall?id=log4cxx

It's scanning a bunch of files that we don't care about at the moment(javascript/css), so I'll need to take a look at that.

PR #107, opened 2022-01-28
boost-fallback: only check for boost if really needed

pseiderer:
- only check for boost if really needed (in case of legacy c++ standard < c++17)

Signed-off-by: Peter Seiderer <ps.report@gmx.net>
---
Notes:

  - fixes the following buildroot configure/compile failure in case
    a only partly available boost is detected (but not needed as a
    c++17 capable compiler is available):

      -- Found Boost: .../host/x86_64-buildroot-linux-uclibc/sysroot/usr/include (found version "1.78.0") found components: thread chrono missing components: date_time atomic
      CMake Error at .../build/log4cxx-0.12.0/boost-fallback-compile-tests/CMakeFiles/CMakeTmp/CMakeLists.txt:19 (add_executable):
        Target "cmTC_aac37" links to target "Boost::date_time" but the target was
        not found.  Perhaps a find_package() call is missing for an IMPORTED
        target, or an ALIAS target is missing?

      CMake Error at .../build/log4cxx-0.12.0/boost-fallback-compile-tests/CMakeFiles/CMakeTmp/CMakeLists.txt:19 (add_executable):
        Target "cmTC_aac37" links to target "Boost::atomic" but the target was not
        found.  Perhaps a find_package() call is missing for an IMPORTED target, or
        an ALIAS target is missing?

    See http://lists.busybox.net/pipermail/buildroot/2022-January/635042.html

### DIFF
--- a/src/cmake/boost-fallback/boost-fallback.cmake
+++ b/src/cmake/boost-fallback/boost-fallback.cmake
@@ -55,6 +55,8 @@ try_compile(STD_SHARED_PTR_FOUND "${CMAKE_BINARY_DIR}/boost-fallback-compile-tes
 try_compile(STD_ATOMIC_FOUND "${CMAKE_BINARY_DIR}/boost-fallback-compile-tests"
     "${CMAKE_CURRENT_LIST_DIR}/test-stdatomic.cpp")
 
+# search for boost only in case needed for legacy c++ standard < c++17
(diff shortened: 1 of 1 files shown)


rm5248:
I'm surprised that it failed - I thought that the try_compile tests would not cause it to completely fail.  Maybe I just don't understand enough about how it works.

rm5248:
I'm looking at an issue([LOGCXX-551](https://issues.apache.org/jira/browse/LOGCXX-551)) that has come up due to this fix - how did this situation happen in the first place?  This looks a lot like the boost installation is very broken, as I've been unable to reproduce the issue, and the cmake FindBoost should properly account for the transitive boost dependencies.

rm5248:
Wait, I've figured it out - your build is using cmake 3.16, which is not smart enough to know about boost 1.78 and thus does not take in the transitive dependencies required.

PR #108, opened 2022-03-19
LOGCXX-550 The thread name should now be able to be logged

rm5248:


ams-tschoening (approved):
For some reason `timebasedrollingtest` failed once for me, but not on subsequent 5 executions. So things seem to be fine.

rm5248:
> For some reason `timebasedrollingtest` failed once for me, but not on subsequent 5 executions. So things seem to be fine.

That is likely due to how that test works(it sleeps a lot during the test).  It's very unreliable on the github builds.  LOGCXX-527 tracks the fix for that whenever I get around to it.

PR #110, opened 2022-05-05
Fixed version number typo

jstsddg:
I was confused while searching for the newest version and could not find v0.13.1.
To stop others from searching as well, I fixed the typo. Thanks for this great library!

The version number was mistakenly displayed as `0.13.1` instead of `0.13.0`.


rm5248:
thanks for the fix!

PR #111, opened 2022-06-02
Logcxx 556

rm5248:
Fix a few issues with the syslog appender:

- Add default layout if none was provided(segfault when using default constructor)
- Check for minimum message size
- Added test

### DIFF
--- a/src/main/cpp/syslogappender.cpp
+++ b/src/main/cpp/syslogappender.cpp
@@ -327,12 +327,7 @@ void SyslogAppender::append(const spi::LoggingEventPtr& event, Pool& p)
 
 		while ( start != msg.end() )
 		{
-			LogString::iterator end;// = start + maxMessageLength - 12;
-
-//			if ( end > msg.end() )
-//			{
-//				end = msg.end();
-//			}
+			LogString::iterator end;
 
 			if( msg.size() > (maxMessageLength - MINIMUM_MESSAGE_SIZE) ) {
 				end = msg.end();
--- a/src/main/cpp/syslogappender.cpp
+++ b/src/main/cpp/syslogappender.cpp
@@ -70,7 +70,7 @@ SyslogAppender::SyslogAppender()
 	: syslogFacility(LOG_USER), facilityPrinting(false), sw(0), maxMessageLength(1024)
 {
 	this->initSyslogFacilityStr();
-	this->layout = LayoutPtr(new PatternLayout("%m"));
+	this->layout = LayoutPtr(new PatternLayout(LOG4CXX_STR("%m")));
 }
 
 SyslogAppender::SyslogAppender(const LayoutPtr& layout1,
--- a/src/main/cpp/syslogappender.cpp
+++ b/src/main/cpp/syslogappender.cpp
@@ -23,6 +23,7 @@
 #include <log4cxx/level.h>
 #include <log4cxx/helpers/transcoder.h>
 #include <log4cxx/helpers/optionconverter.h>
+#include <log4cxx/patternlayout.h>
 #if !defined(LOG4CXX)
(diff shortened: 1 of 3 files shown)


src/main/cpp/syslogappender.cpp
```diff
@@ -326,11 +327,17 @@ void SyslogAppender::append(const spi::LoggingEventPtr& event, Pool& p)
...
+			LogString::iterator end;// = start + maxMessageLength - 12;
+
+//			if ( end > msg.end() )
+//			{
+//				end = msg.end();
+//			}
 
-			if ( end > msg.end() )
-			{
+			if( msg.size() > (maxMessageLength - MINIMUM_MESSAGE_SIZE) ) {
```
ams-tschoening, line 332: `maxMessageLength - MINIMUM_MESSAGE_SIZE` looks like some sort of `msgWindowLen[gth]`? Doesn't seem to change, so might be maintained on instance level, calculated with `setMaxMessageLength`. Makes code easier to understand in my opinion, like with removing `12` in favour of some name.

ams-tschoening, line 332: `msg.size() > (maxMessageLength - MINIMUM_MESSAGE_SIZE)` seems to not change within the loop, so put it above to declaration of `end`?

src/main/cpp/syslogappender.cpp
```diff
@@ -326,11 +327,12 @@ void SyslogAppender::append(const spi::LoggingEventPtr& event, Pool& p)
 
 		while ( start != msg.end() )
 		{
-			LogString::iterator end = start + maxMessageLength - 12;
+			LogString::iterator end;
 
-			if ( end > msg.end() )
-			{
+			if( msg.size() > (maxMessageLength - MINIMUM_MESSAGE_SIZE) ) {
```
ams-tschoening, line 332: Line 349 contains the following line which might be changed to using `MINIMUM_MESSAGE_SIZE` instead:

    char buf[12];

ams-tschoening (commented):
Some minor suggestions...

src/main/cpp/syslogappender.cpp
rm5248: Actually, that brings up a different point: what I have doesn't work in the first place - what should be happening is that you should be splitting it up into N messages.  The issue indicates that windows does not like to have an iterator past the end of the string(which this code does).  So it really just needs to calculate the N chunks at the beginning and do the correct number of iterations to split up the string properly.

src/main/cpp/syslogappender.cpp
```diff
@@ -322,33 +323,37 @@ void SyslogAppender::append(const spi::LoggingEventPtr& event, Pool& p)
 
 	if ( msg.size() > maxMessageLength )
 	{
+		int numberOfFullChunks = msg.size() / maxMessageLength;
 		LogString::iterator start = msg.begin();
+		LogString::iterator end = msg.begin() + maxMessageLength;
```
ams-tschoening, line 328: > // Note: MSVC/Windows does not like it when you seek past the end of the string

Doesn't that comment apply for this line as well? Shouldn't there be a check for number of chunks aka the length of the string being greater than `maxMessageLength`?

    LogString::iterator end = numberOfFullChunks > 0|1 ? (msg.begin() + maxMessageLength) : msg.end();

Not sure about the case when message length and `maxMessageLength` are equal.

rm5248: Unless I have something wrong with my math, that line of code should only be relevant if the message is larger than the maximum message size(in which case it needs to split it up into several chunks).

ams-tschoening: I didn't recognize the `if` before. :-)

PR #114, opened 2022-08-14
LOGCXX-549

rm5248:
Make networking optional and abstract so that different network backends can be used in the future.


CMakeLists.txt
```diff
@@ -257,3 +258,17 @@ message(STATUS "  zip ............................. : ${ZIP_APP}")
...
+
+message(STATUS "Available appenders:")
+message(STATUS "  Async Appender .................. : ON")
+message(STATUS "  ODBC Appender ................... : ${HAS_ODBC}")
+message(STATUS "  SMTP Appender ................... : ${HAS_LIBESMTP}")
+message(STATUS "  XMLSocketAppender ............... : ${LOG4CXX_NETWORKING_SUPPORT}")
+message(STATUS "  SocketHubAppender ............... : ${LOG4CXX_NETWORKING_SUPPORT}")
+message(STATUS "  SyslogAppender .................. : ${LOG4CXX_NETWORKING_SUPPORT}")
+message(STATUS "  TelnetAppender .................. : ${LOG4CXX_NETWORKING_SUPPORT}")
+message(STATUS "  NTEventLogAppender .............. : ${LOG4CXX_NETWORKING_SUPPORT}")
```
ams-tschoening, line 270: Should `NTEventLogAppender` really be considered being part of networking? Its header is placed in `nt` instead of `net` and does work without networking in Windows AFAIK.

rm5248: You are absolutely right, I got a little excited with the copy/paste there.

PR #116, opened 2022-08-20
LOGCXX-527 take 2

rm5248:


rm5248:
I am so confused as to what is going on with git here

PR #119, opened 2022-08-28
LOGCXX-527: mock the clock for tests

rm5248:
Mock the clock for unit tests to make them more reliable and configurable.
This also cleans up some APR calls to make us less dependent on APR and
more on standard C++ features

Some former assumptions didn't hold anymore. Some tests use generic file
names and if those exist during rollover, their last write time is used as
new file name. That last write time might be arbitrary in the past and
therefore broke existing checks about expected file names. Hence, those files have been deleted.

That deletion was based on some naming convention and the number of the
executed tests. But some tests simply didn't follow that convention and
because some tests where commented, the expected numbers didn't work as
well. So I've changed the implementation so that each test method needs
to delete the file, as only each method knows which name is used.

Additionally, I centralized things into an individual directory, as that
is done for the corresponding witnesses as well already. Some witness
file contents needed to be adjusted to what I think is the new behaviour,
but not too sure who is correct, the former witness or the test implementation.


rm5248:
@ams-tschoening I squashed our two commits since otherwise git didn't want to merge it in(I think because the commit was reverted later on).

PR #120, opened 2022-09-03
Stop all file watch dog instances in LogManager::shutdown()

swebb2066:
This PR aims to address LOG4CXX-451.

Alos included is support for automatic start of a watchdog when using automatic configuration by providing a positive integer value in the LOG4CXX_CONFIGURATION_WATCH_SECONDS environment variable.


swebb2066:
Wrong branch

PR #122, opened 2022-09-09
APR termination must be after destruction of all log4cxx objects

swebb2066:
As the APRInitializer class has member variables whose destructors call apr_* functions, apr_terminate() cannot be called the the APRInitializer class destructor.

### DIFF
--- a/src/main/cpp/appenderattachableimpl.cpp
+++ b/src/main/cpp/appenderattachableimpl.cpp
@@ -32,7 +32,7 @@ struct AppenderAttachableImpl::priv_data
 {
 	/** Array of appenders. */
 	AppenderList  appenderList;
-	mutable std::mutex m_mutex;
+	mutable std::recursive_mutex m_mutex;
 };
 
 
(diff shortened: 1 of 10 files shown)


rm5248:
This looks good to me.

I was concerned about the comment on adding/removing appenders if there is a problem and I added a new test for that.  It doesn't appear to be an issue.  commit: https://github.com/apache/logging-log4cxx/commit/cb96ae91fd73e94bcf34a06077e991b9ae477fdc

Note that when I first ran your branch, it did segfault on the multithread test.  As long as it's not 100% of the time, I'm not too concerned about it.

swebb2066:
A stack trace of a crash in mutithreadtest (after hundreds of instatiations) - see [LOGCXX-532](https://issues.apache.org/jira/projects/LOGCXX/issues/LOGCXX-532?filter=allopenissues) - indicates the cause of the crash is likely to be independent of log4cxx.

swebb2066:
You were correct to be concerned about the comment on adding/removing appenders.

A FallbackErrorHandler that holds the same logger that caused the fall back would have resulted in a deadlock.

The latest commit should fix this.

PR #123, opened 2022-09-10
Fix sonic cloud bugs

swebb2066:


rm5248:
Does the `InterruptedException` class still exist?  That should probably just be removed.

PR #125, opened 2022-09-10
Remove asyncappender_nonblocking.cpp

swebb2066:
asyncappender_nonblocking.cpp was a copy of asyncappender.cpp as at 2018 taken prior to improvements to asyncappender.cpp

The master version of asyncappender_nonblocking.cpp has compileation errors.

I doubt it has ever been used


rm5248:
Looks good to me, agree that it should be removed.

Side note: I don't think the `$<IF:$<BOOL:LOG4CXX_BLOCKING_ASYNC_APPENDER...` works.  When I tried to set the option it didn't seem to change anything.

PR #126, opened 2022-09-11
Prevent a segmentation fault during initialization

swebb2066:
This PR prevents a runtime crash when using a staticly linked log4cxx

### DIFF
--- a/src/main/cpp/date.cpp
+++ b/src/main/cpp/date.cpp
@@ -27,10 +27,11 @@ using namespace log4cxx::helpers;
 
 IMPLEMENT_LOG4CXX_OBJECT(Date)
 
+namespace {
+Date::GetCurrentTimeFn getCurrentTimeFn = 0;
+}
 
-Date::GetCurrentTimeFn Date::getCurrentTimeFn = Date::getCurrentTimeStd;
-
-Date::Date() : time(getCurrentTimeFn())
+Date::Date() : time(currentTime())
 {
 }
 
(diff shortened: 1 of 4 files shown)


rm5248:
Is this not currently caught by the static windows build that we do?

swebb2066:
master does not have that change to workflows

rm5248:
next_stable does have that change though.  It looks like the build is performed, but the since the tests aren't run with the static build any runtime error would not show up.

I think that we want to split out the build and test of a static build into a separate workflow so we can catch this if it fails again.

Otherwise this looks good to me.

LeeGDavis:
I think I'm running into this. We're in the process of upgrading from 0.11 to 0.13 and encountering a segfault on initialization. Is there anyway I can test this with my application using vcpkg and report back? Also If I shouldn't be posting here, please advise.

rm5248:
The best place to get help would be to send a message to the mailing list(log4cxx-user@logging.apache.org)

However if you are using vcpkg, you should just be able to create an overlay port to override the version that is in vcpkg, or modify vcpkg locally on your side.  This may help: https://devblogs.microsoft.com/cppblog/registries-bring-your-own-libraries-to-vcpkg/

LeeGDavis:
Thank you for the links @rm5248 !  I'll take a look to the mailing list moving forward too.

LeeGDavis:
That worked for us.  There was one problem with the portfile.cmake. I had to revert recent threadutility.cc changes and nuke linux.patch. Otherwise we are no longer segfaulting with this change.  We have yet to test all the functionality, but I look forward to getting our hands on this update.  Thanks again!

PR #127, opened 2022-09-11
Prevent compilation error when compiling fmt example against fmt v9.0.0

swebb2066:


rm5248:
Does the `#include <fmt/ostream.h>` not let this work?  I know it did work, but maybe it changed - the [documentation](https://fmt.dev/latest/api.html#std-ostream-support) seems different.

I'd like to keep it using `operator<<` if possible, since that means that the example would work with both `LOG4CXX_DEBUG()` and `LOG4CXX_DEBUG_FMT()`, although if we want to create a new example showing the usages that would be find as well.

PR #128, opened 2022-09-11
Change AsynchAppender to not switch to synchronous behaviour

swebb2066:
Allow a FallbackErrorHandler to replace the bad appender when an attached appender throws an exception.

Resolves LOGCXX-432


rm5248:
Looks good to me, that should fix the problem as described.

PR #131, opened 2022-09-17
Google benchmark option

swebb2066:
google benchmark has a lot of tool support (e.g. comparison) and (hopefully) less variance in run-time estimates (seems to still be 10% on an idle Ubuntu logging using patterns)

### DIFF
--- a/src/test/cpp/CMakeLists.txt
+++ b/src/test/cpp/CMakeLists.txt
@@ -81,6 +81,7 @@ add_subdirectory(spi)
 add_subdirectory(varia)
 add_subdirectory(xml)
 add_subdirectory(throughput)
+add_subdirectory(benchmark)
 
 # Note: we need to include the APR DLLs on our path so that the tests will run.
 # The way that CMake sets the environment is that it actually generates a secondary file,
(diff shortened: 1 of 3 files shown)


swebb2066:
Output example:
`----------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations
----------------------------------------------------------------------------------------
Logging disabled trace                              7.00 ns         7.00 ns     97609134
Logging disabled trace/threads:8                    2.01 ns         15.6 ns     50647640
Logging disabled debug                              7.15 ns         7.15 ns     97977587
Logging info static string                           662 ns          662 ns      1043254
Logging enabled debug static string                  653 ns          653 ns      1072875
Logging enabled trace static string                  659 ns          659 ns      1063702
Logging static string with FMT                       654 ns          654 ns      1059184
Logging int value with FMT                           720 ns          720 ns      1005913
Logging int value with FMT/threads:8                 996 ns         7442 ns        94872
Logging int value with std::ostream                 1227 ns         1227 ns       570771
Logging int value with std::ostream/threads:8       1218 ns         9420 ns        74304
NoFormat pattern: %m%n                              1242 ns         1242 ns       573767
DateOnly pattern: [%d] %m%n                         1240 ns         1240 ns       568791
DateClassLevel pattern: [%d] [%c] [%p] %m%n         1240 ns         1240 ns       570043
s`

PR #133, opened 2022-09-21
prefer pkg-config over apu-1-config executable

DerDakon:
When calling pkg-config paths that are relative to a sysroot will automatically get the correct prefix, so try using that first.


swebb2066:
Does the same issue arise with APR?

DerDakon:
Yes and no. Yes, it does, but I did not notice it because I had to patch apr-1-config to be able to build apr-util at all. Because cross-building that is it's own problem as it doesn't look into .pc files either.

Will add a follow-up patch to get that right as well.

src/cmake/FindAPR-Util.cmake
```diff
@@ -27,30 +27,40 @@ macro(_apu_invoke _varname)
...
-    if (APU_STATIC OR NOT BUILD_SHARED_LIBS)
-      set(APR_UTIL_COMPILE_DEFINITIONS APU_DECLARE_STATIC)
-      find_library(APR_UTIL_LIBRARIES NAMES aprutil-1)
-    else()
-      find_library(APR_UTIL_LIBRARIES NAMES libaprutil-1)
-      find_program(APR_UTIL_DLL libaprutil-1.dll)
-    endif()
+  find_program(APR_UTIL_CONFIG_EXECUTABLE
+      apu-1-config
+      PATHS /usr/local/opt/apr-util/bin    $ENV{ProgramFiles}/apr-util/bin
```
rm5248, line 41: Is there a reason to remove the other PATHS from here?  If we need to fallback to using apu-1-config(and apr-1-config) it seems like we should check on the normal PATH directories and not limit ourselves to just `/usr/local/opt/...`.

rm5248 (commented):
Using pkg-config should probably be preferred but it doesn't seem to work well with OSX for some reason.  Is this a problem we can detect and fall back to using `apu-1-config` if the pkgconfig fails?

src/cmake/FindAPR-Util.cmake
DerDakon: Yes, because ```find_program()``` will look there anyway. These PATHS are just additional search paths.

DerDakon:
> Using pkg-config should probably be preferred but it doesn't seem to work well with OSX for some reason. Is this a problem we can detect and fall back to using `apu-1-config` if the pkgconfig fails?

I will try using just the variables from pkg-config instead of the target. This is in no place worse than we had before, but it can't transport things like additional compile flags that the ```PkgConfig::<target>``` variant can. I would still like to find out where the thing actually is on MacOS, just for clarity. But what is clear so far is that the .pc file there contains at least one completely wrong entry.

PR #135, opened 2022-09-26
Reduce disabled logger overhead

swebb2066:
Prior to the PR, next_stable  testing disabled logger performance was 20% worse than master (13.0).

By adding a cached level to Logger,  testing disabled logger performance improves on master by 50%.

Automatic repository configuration is now explicit part (autoConfigure() function) of the LoggerRepository interface, and LogManager is responsible for calling autoConfigure(). This is required to have correct state for the cached level thresholds.

### DIFF
--- a/src/main/cpp/hierarchy.cpp
+++ b/src/main/cpp/hierarchy.cpp
@@ -238,6 +238,7 @@ LoggerPtr Hierarchy::getLogger(const LogString& name)
 LoggerPtr Hierarchy::getLogger(const LogString& name,
 	const spi::LoggerFactoryPtr& factory)
 {
+	auto root = getRootLogger();
 	std::unique_lock<std::mutex> lock(m_priv->mutex);
 
 	LoggerMap::iterator it = m_priv->loggers.find(name);
@@ -261,7 +262,7 @@ LoggerPtr Hierarchy::getLogger(const LogString& name,
 			m_priv->provisionNodes.erase(it2);
 		}
 
-		updateParents(logger);
+		updateParents(logger, root);
 		result = logger;
 	}
 	return result;
@@ -370,7 +371,7 @@ void Hierarchy::shutdownInternal()
 	}
 }
 
-void Hierarchy::updateParents(LoggerPtr logger)
+void Hierarchy::updateParents(const LoggerPtr& logger, const LoggerPtr& root)
 {
 	const LogString name(logger->getName());
 	size_t length = name.size();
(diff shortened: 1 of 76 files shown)


rm5248:
I'll need to test this out sometime in the next few days, but that looks good to me.  One thing that I would like to double-check is to make sure that re-configuring works correctly at runtime(e.g. calling DOMConfigurator::configure), since it looks like the reparenting and level checking has changed slightly.

Are there any other commonly accessed fields that may need to be private?  In this case adding it as a private member variable instead of in the private class is reasonable to do for performance purposes.

rm5248:
So after taking a look at this, what I was afraid of(not propagating the level to sub-loggers) is true.  For example, if we have loggers `com` and `com.example` and we reconfigure to set the logger `com` to be `TRACE` for example, the `com.example` logger does not also get this level.  This is assuming that the log configuration is coming from an XML file, although I think this would also be the case if you manually set the log levels.  If you load an XML file with `com` set to `TRACE` at the beginning of execution, you will see the `com` and `com.example` messages output.

This currently works fine in next_stable.  I'll add a test for this feature soon, since it is something that I depend on often.  There is a related issue, if you remove a logger from the configuration file and reload the file the logger will retain its currently set level.

Unfortunately I don't really see an easy way of fixing this.  We could reconfigure everything once we load the file, but that leads to a less than ideal situation of if a user is manually setting the log level they would now have to do an extra call at the end to have all of the loggers cache their new threshold level.

swebb2066:
** Unfortunately I don't really see an easy way of fixing this.

I think the logger can ask the Hierarchy to update its children - Hierachy has all the loggers.

The principle here is that work should br done in infrequent events (e.g. changing a level) in preference to burdening every frequent event.

src/main/cpp/hierarchy.cpp
```diff
@@ -433,6 +433,22 @@ void Hierarchy::updateChildren(ProvisionNode& pn, const LoggerPtr& logger)
...
+    
+}
+
+void Hierarchy::updateChildren(const Logger* parent)
+{
+	for (auto& item : m_priv->loggers)
+	{
+		for (auto l = item.second; l; l = l->getParent())
+		{
+			if (l->getParent().get() == parent)
```
rm5248, line 445: I'm pretty sure this is going to only go one level deep though?  so `com.example` would get set but not `com.example.foo`

rm5248: Oh yes you're right, I misread what the code was doing.

swebb2066:
**Are there any other commonly accessed fields that may need to be private?
Not that I can see,

PR #137, opened 2022-09-30
Document DefaultConfigurator actual behaviour

swebb2066:
This PR also allows the user to specify the configuration file and the watch seconds in program code.

This makes automatic configuration more useful by allowing the configuration file (and watchdog seconds) to be set in a user's getLogger().

For example:

```
log4cxx::LoggerPtr MyLogManager::getLogger(const log4cxx::LogString& name)
{
  static std::once_flag initialize_log4cxx;
  std::call_once(initialize_log4cxx, DefaultConfigurator::setConfigurationFileName, LOG4CXX_STR("myApp.xml") );
  return log4cxx::LogManager::getLogger(name);
}
```


src/main/include/log4cxx/defaultconfigurator.h
```diff
@@ -39,19 +35,39 @@ class LOG4CXX_EXPORT DefaultConfigurator
...
-		prints to <code>stdout</code> to the root logger.*/
-		static void configure(log4cxx::spi::LoggerRepositoryPtr repository);
+		Configure \c repository.
+
+		If the configuration file name has not been provided by a call to setConfigurationFileName(),
+		the environment variables 'LOG4CXX_CONFIGURATION' and 'log4j.configuration' are examined.
+
+		If the configuration file name has not been provided by any of the previous approaches
+		the current directory is examined for a file named either
+		"log4cxx.xml", "log4cxx.properties", "log4j.xml" or "log4j.properties".
```
rm5248, line 45: what happens if this fails?  Does it still fall back to the old way of using the `PatternLayout`?

rm5248:
The latest failure on OSX may be a timing issue.  If there's a way to avoid the sleep in the test we should probably do that, otherwise that will likely continue to intermittently fail.

For `DefaultConfigurator::setConfigurationWatchSeconds` should this be `std::chrono` instead of just raw int?

src/main/include/log4cxx/defaultconfigurator.h
swebb2066: You will see the unhelpful "Please initialize the log4cxx system properly".

It also writes a LogLog::debug message.

swebb2066:
* For DefaultConfigurator::setConfigurationWatchSeconds should this be std::chrono instead of just raw int?

The simplicity of a raw int (seconds) appeals to me.

swebb2066:
** The latest failure on OSX may be a timing issue.

The file append modification does not seem to change the `mtime` returned by stat.

swebb2066:
**  If there's a way to avoid the sleep in the test we should probably do that.
Sleeping longer makes the test more robust by ensuring the creation time and modification time are different.

PR #140, opened 2022-10-07
Prefer a file named like the executable for configuration data.

swebb2066:
I believe naming the configuration file like the executable this is a very common use case.

This PR supports testing (where the configuration file is in the current working directory) and the shipped product (where the configuration file is in the same directory as the executable).


src/main/cpp/defaultconfigurator.cpp
```diff
@@ -60,17 +131,32 @@ void DefaultConfigurator::configure(LoggerRepositoryPtr repository)
 
 	if (configurationFileName.empty())
 	{
-		const char* names[] = { "log4cxx.xml", "log4cxx.properties", "log4j.xml", "log4j.properties", 0 };
```
ams-tschoening, line 63: How about not removing those names and instead use them additionally as suffix to the executed binary? I'm following this naming scheme and am somewhat sure I've read about it elsewhere as well in the past few years:

```
some.exe
some.exe.config.xml
some.exe.log4cxx.xml
some.dll
some.dll.[...]
```

Microsoft is using something similar:

```
VSPerfReport.exe.manifest
testhost.net48.arm64.exe.config
```

Having only the extension XML or else to the binary name is too broad in my opinion, instead especially `.xml`, `.json`, `.yaml` etc. is used for app configs often.

swebb2066: I will have to abandon this PR as I do not know how to make it work on Macos.

I am thinking I will put the file name selection code in an example MyLogManager.cpp instead.

swebb2066:
There are to many possibilities for configuration file names and operating systems

PR #148, opened 2022-11-07
Reduce std::ostream logging overhead

swebb2066:
This PR reduces the overhead of logging values through `operator<<(std::ostream&` by using a thread local std::basic_ostringstream where available.

It retains the static string logging optimization,  thereby resolving the issue with the previous attempt ([LOGCXX-500](https://issues.apache.org/jira/browse/LOGCXX-500)) to introduce this optimization.

### DIFF
--- a/src/main/cpp/messagebuffer.cpp
+++ b/src/main/cpp/messagebuffer.cpp
@@ -22,21 +22,12 @@
 #endif
 #include <log4cxx/helpers/messagebuffer.h>
 #include <log4cxx/helpers/transcoder.h>
+#if !defined(LOG4CXX)
+	#define LOG4CXX 1
+#endif
+#include <log4cxx/private/log4cxx_private.h>
 
 using namespace log4cxx::helpers;
-
-static bool gMessageBufferUseStaticStream = false;
-
-namespace log4cxx
-{
-namespace helpers
-{
-void MessageBufferUseStaticStream()
-{
-	gMessageBufferUseStaticStream = true;
-}
-}
-}
 
 template <typename T>
 void ResetStream(std::basic_ostringstream<T>& stream)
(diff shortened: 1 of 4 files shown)


src/main/cpp/messagebuffer.cpp
```diff
@@ -149,10 +118,14 @@ CharMessageBuffer& CharMessageBuffer::operator<<(const char msg)
...
-	if (m_priv->stream == 0)
+	if (!m_priv->stream)
 	{
+#if LOG4CXX_HAS_THREAD_LOCAL
+		thread_local static std::basic_ostringstream<char> sStream;
+		m_priv->stream = &sStream;
+#else
 		m_priv->stream = new std::basic_ostringstream<char>();
-
+#endif
```
rm5248, line 128: At this point I'm not aware of any compiler that supports C++11 but not `thread_local`, would it make sense to simply make `m_priv->stream` a `thread_local` variable instead of holding a pointer?  I could be wrong here, but it looks like if two threads are logging at the same time they could both get the same `ostringstream`, thus defeating the purpose of the `thread_local`.

swebb2066: > would it make sense to simply make `m_priv->stream` a `thread_local `variable instead of holding a pointer

The `m_priv->stream` must be zero when logging only `std::string` or `char* ` to implement the static string logging optimization.

> looks like if two threads are logging at the same time they could both get the same ostringstream

The `thread_local ` implementation should ensure the sSteam obtained by each thread will be different, and the `m_priv->stream` pointer life-time is the `MessageBuffer` lifetime. The only way two threads could share the same `ostringstream` is if the user provides a `MessageBuffer` object reference to the other thread. `MessageBuffer` is not intended to be used outside LOG4CXX_* macros.

PR #159, opened 2022-12-12
LOGCXX-560 logging in a static initializer

swebb2066:
Change to documentation so the configuration example creates a system in which logging in a static initializer works.

Also make auto-configuration example (using DefaultConfigurator) similar.


rm5248:
I see that you updated the example in the usage page.  Are the changes that you put in the usage.md file because of the logging in a static initializer?

The reason that I'm asking is because I wonder if this is getting too complicated for an introductory example; the `com::foo::getLogger` function is not needed for basic applications.  I have never had a need to log in a static initializer(although I'm sure some people have).  Would it make sense to make a new section in the usage page(or create a new page) that goes over what is required for logging in a static initializer?

swebb2066:
I changed index.md because I believe wrapping `getLogger `is a best-practice technique. It will improve user code maintainability, compared to initializing log4cxx in main.

Why do you think have a wrapper fumction for `getLogger` is complicated? Is it because you think a function `local static struct `is too complicated?

I would expect level of software engineer using log4cxx (which is really only appropriate for large applications) should be able to handle a `local static struct` technique.

rm5248:
My thinking is more along the lines of a first time user, having a small
standalone example is very useful. Both spdlog and boost log have simple
examples that are only about 10-20 lines long.

I don’t think it’s a bad example(it is needed), I just think it’s a little
complicated for the first example. Some of it depends on how you use the
library - I always initialize the logging in main, because I can switch it
around with a command line parameter.  I also never log in a static
initializer so there’s no need for me to load the config file before main
starts.

rm5248:
To expand on my previous comment, I would propose that the user guide look something like the following, basically showing the two options on how to use.  The first example is pretty much what exists today, but more simplified.  The second example is what you have here.

For the first-time user the idea with the first example is that because this is only 30 lines, it is very easy to copy and paste into a scratch project to test things out with.

-----
# Configuration
## A Simple Example
In order to start using Log4cxx, a simple example program is shown below.  This program does nothing useful, but it shows the basics of how to start using Log4cxx.  Using the BasicConfigurator class, we are able to quickly configure the library to send data to standard error at the INFO level.
```
   // include log4cxx header files.
    #include "log4cxx/logger.h"
    #include "log4cxx/basicconfigurator.h"
     
    using namespace log4cxx;
     
    static LoggerPtr logger(Logger::getLogger("MyApp"));

void foo(){
  // Get a logger that is a child of the statically declared logger
  LoggerPtr fooLogger(Logger::getLogger("MyApp.foo");
  LOG4CXX_TRACE(fooLogger, "Doing foo at trace level");
LOG4CXX_DEBUG(fooLogger, "Doing foo at debug level");
LOG4CXX_INFO(fooLogger, "Doing foo at info level");
LOG4CXX_WARN(fooLogger, "Doing foo at warn level");
LOG4CXX_ERROR(fooLogger, "Doing foo at error level");
LOG4CXX_FATAL(fooLogger, "Doing foo at fatal level");
}
     
    int main(int argc, char **argv)
    {
    	int result = EXIT_SUCCESS;
    // Set up a simple configuration that logs on the console.
    BasicConfigurator::configure();
   
    LOG4CXX_INFO(logger, "Entering application.");
    foo();
    LOG4CXX_INFO(logger, "Exiting application.");
     
    	return result;
    }
```
The above application does nothing useful except to show how to initialize logging with the BasicConfigurator and do logging with different loggers.  Note that file based configurations are also possible - see DOMConfigurator and PropertiesConfigurator.

This example also has some limitations in how logging is initialized:

- Any logging in static initializers will be lost, as the logging is not initialized before main() starts
- (some other limitation??)

## Configuration for use with static initialization

(your example here)

There are some advantages to this approach:

- You are able to log in static initializers when configuring through code if the default initialization procedure is not appropriate
- (some other advantage that I don't know about??)

PR #164, opened 2022-12-17
LOGCXX-568 Cleanup if GZip fails

rm5248:
Cleanup 0-sized gzip files if rollover fails.  Added a new option to determine what to do if there is a failure, to make the rollover dumb.

This also implements the long-standing PR #62, adding that fix in as an option.  The best way to fix this is to use the `FallbackErrorHandler`, however this fix is easier to configure.

### DIFF
--- a/src/main/include/log4cxx/rolling/gzcompressaction.h
+++ b/src/main/include/log4cxx/rolling/gzcompressaction.h
@@ -52,6 +52,17 @@ class GZCompressAction : public Action
 		 */
 		bool execute(log4cxx::helpers::Pool& pool) const override;
 
+		/**
+		 * Set to true to throw an IOException on a fork failure.  By default, this
+		 * is true.  When an IOException is thrown, this will automatically cause the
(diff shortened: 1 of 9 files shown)


src/main/include/log4cxx/rolling/gzcompressaction.h
```diff
@@ -52,6 +52,17 @@ class GZCompressAction : public Action
 		 */
 		bool execute(log4cxx::helpers::Pool& pool) const override;
 
+		/**
+		 * Set to true to throw an IOException on a fork failure.  By default, this
+		 * is true.  When an IOException is thrown, this will automatically cause the
+		 * error handler to be called(which is the recommended way of handling this
+		 * problem).  By setting this to true, the GZCompressAction effectively
```
nturner, line 59: Did you mean "By setting this to false, ..."?

rm5248: whoops, thanks for noticing!  fixed.

PR #170, opened 2022-12-22
LOGCXX-567

rm5248:
A simpler implementation of stacktrace support that simply puts the stacktrace into the MDC before the message is logged.


swebb2066:
I really do not think a stack trace on every log statement is anywhere near being a good idea.

rm5248:
> I really do not think a stack trace on every log statement is anywhere near being a good idea.

I agree; that's why this is just a convenience setting that you need to enable explicitly by defining `LOG4CXX_ENABLE_STACKTRACE`.  If this macro is not defined nothing happens.

swebb2066:
I suggest separate macros instead of a change to the existing macro.

I suggest LOG4CXX_WARN_WITH_STACK and LOG4CXX_ERROR_WITH_STACK would be sufficient.

rm5248:
The reason I didn't make separate macros is that I don't want to change log statements if I now want a stack trace.  The intention is that you simply `#define LOG4CXX_ENABLE_STACKTRACE` at the top of your file, so that within that file you will get stacktraces as part of your log statements(this is similar to the Linux kernel `DEBUG` macro, where if it is not defined debug log statements will be compiled out).  The stack traces are not very useful if you're not building a debug build anyway, because you won't get the method names.

So in my mind there are a few ways we could do this:

1. Keep as-is, but add new macros that unconditionally add in the stacktraces
2. Add in a new macro that prints out a stack trace(something like `LOG4CXX_STACKTRACE(logger)`)
3. Create some new macros that determine what level(s) the stacktrace is inserted into the MDC at(similar to the `LOG4CXX_THRESHOLD` macro)
4. Ignore this entirely and simply document some examples of how you might use Boost stacktrace to generate a stacktrace

Originally I had it so that the stacktrace was generated by internal log4cxx code, but the problem with that is that by the time you create the stacktrace you're already 10-12 layers in, so the stacktrace is not very useful - it really needs to start at the log statement.

ams-tschoening:
> I suggest LOG4CXX_WARN_WITH_STACK and LOG4CXX_ERROR_WITH_STACK would be sufficient.

Those macros read like they will always include stacktraces in the output to me, which is not the case, things additionally depend on the configured pattern. Why care at all when some dev wants to have stacktraces, if it's for DEBUG or ERROR?

> The reason I didn't make separate macros is that I don't want to change log statements if I now want a stack trace.

This argument makes sense to me and the naming of `*_ENABLE_*` doesn't even imply the stacktrace is always embedded into the log message, but only its availability.

swebb2066:
> The reason I didn't make separate macros is that I don't want to change log statements if I now want a stack trace.

What situation you are thinking of where adding a stack-trace to pre-existing log statements would be useful?

swebb2066:
Given that C++23 paper [A Proposal to add stacktrace library](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0881r3.html) is motivated by **assertions can't describe the whole picture of a bug**, perhaps there should be a new LOG4CXX_ASSERT macro

rm5248:
> > The reason I didn't make separate macros is that I don't want to change log statements if I now want a stack trace.
> 
> What situation you are thinking of where adding a stack-trace to pre-existing log statements would be useful?

It hasn't come up for me for a while, but occasionally I do need to know what piece of code is calling a certain function, perhaps because a function is called a number of different places and I don't know how we got there.  This can be easier than stepping through the debugger and hitting the same breakpoint 20 times.

> Given that C++23 paper [A Proposal to add stacktrace library](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0881r3.html) is _only_ motivated by **assertions can't describe the whole picture of a bug**, perhaps the stacktrace should _only_ be added to the LOG4CXX_ASSERT macro.
> 

That sounds reasonable to me; I will add some documentation on how you could add the stacktrace to normal log statements if you want that for whatever reason.

> Then instead of `LOG4CXX_ENABLE_STACKTRACE` the macro would be `LOG4CXX_HAS_STACKTRACE` which is set by when `__cpp_lib_stacktrace` is defined and non-zero or if boost stacktrace is found by cmake.

The advantage of the current implementation is that you don't need to have a new enough C++ standard or have Boost stacktrace installed when compiling log4cxx, which I do like in order to have it as just an optional dependency.

src/main/include/log4cxx/logger.h
```diff
@@ -25,6 +25,10 @@
 #include <log4cxx/helpers/resourcebundle.h>
 #include <log4cxx/helpers/messagebuffer.h>
 
+#if defined(LOG4CXX_ENABLE_STACKTRACE)
```
swebb2066, line 28: We could now avoid the need for LOG4CXX_ENABLE_STACKTRACE using:
```
#ifndef __has_include
#elif __has_include(<stacktrace>)
#include <stacktrace>
#define LOG4CXX_STACKTRACE ::log4cxx::MDC mdc_("stacktrace", LOG4CXX_EOL + std::stacktrace::to_string(boost::stacktrace::stacktrace()));
#elif __has_include(<boost/stacktrace.hpp>)
#include <boost/stacktrace.hpp>
define LOG4CXX_STACKTRACE ::log4cxx::MDC mdc_("stacktrace", LOG4CXX_EOL + boost::stacktrace::to_string(boost::stacktrace::stacktrace()));
#endif
```

src/site/markdown/macros-influencing-log4cxx.md
```diff
@@ -33,3 +33,4 @@ Log4cxx includes), or they may be set globally via your build system.
 | ----- | ----- |
 | LOG4CXX\_THRESHOLD | Used to determine if log messages are compiled in or not.  A higher value causes more messages to be compiled out.  See [removing log statements](usage.html#removing-log-statements) for more information. |
 | LOG4CXX\_DISABLE\_LOCATION\_INFO | Define this macro to disable location information on log statements.  Location information includes the filename, class name, method name, and line number |
+| LOG4CXX\_ENABLE\_STACKTRACE | Define this macro to cause a stacktrace string to be inserted into the MDC with a key of 'stacktrace'.  This requires Boost Stacktrace to be available when compiling your application.  When using the PatternLayout, print out the stacktrace using the `%%X{stacktrace}` specifier. |
```
swebb2066, line 36: If using a pre-C++17 compiler, define ...

src/site/markdown/stacktrace.md
```diff
@@ -0,0 +1,47 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+# Stacktrace Support
+When debugging a code base and an assertion is hit, it is often useful to
+have a stacktrace as part of an assertion in order for you to tell where
+you are in the code to know why it is buggy.  Generating a stacktrace can
+be done with [Boost Stacktrace](https://www.boost.org/doc/libs/1_81_0/doc/html/stacktrace.html).
+
+In order to enable stacktraces when using the `LOG4CXX_ASSERT` family of macros,
```
swebb2066, line 29: ... family of macros with a pre-C++17 compiler,

swebb2066:
Can you add to src/main/include/CMakeLists.txt

`option(LOG4CXX_ENABLE_STACKTRACE "Include a stack trace in the MDC for LOG4CXX_ASSERT output" OFF)`

so the options is available in the cmake GUI?

rm5248:
That wouldn't really do anything, because this is something that needs to be set when you build an application that uses log4cxx, not when you build log4cxx.

The intention is to do something like this in CMake:
```
add_executable(foo foo.cpp)
target_compile_options(foo LOG4CXX_ENABLE_STACKTRACE)
target_link_libraries(foo log4cxx)
```

PR #190, opened 2023-02-04
Prevent priority inversions when logging from multiple threads

swebb2066:
A real-time process with a high priority thread and some low priority threads which both log to the same appender can suffer priority inversions (where the high priority thread is blocked by the IO of the low priorty thread).

See https://bugzilla.redhat.com/show_bug.cgi?id=438484


src/main/include/log4cxx/private/mutex.h
```diff
@@ -0,0 +1,72 @@
...
+{
+
+using AppenderMutexType = std::mutex;
+using AppenderScopeGuard = std::lock_guard<AppenderMutexType>;
+using AppenderScopedLock = std::unique_lock<AppenderMutexType>;
+
+// Change mutex attributes to avoid priority inversion problems
+template <class M>
+	void
+EnablePriorityInheritance(M& m, helpers::Pool& p)
```
rm5248, line 38: It looks like this doesn't do anything at the moment.  From the documentation that I've been reading, it seems that you can only set the mutexattr when creating the mutex in the first place(`pthread_mutex_init`).

swebb2066: > From the documentation that I've been reading
Please post the link -  note that `pthread_mutex_destroy` does mean the it is being created.

rm5248: This was the best thing that I was able to find: https://linux.die.net/man/3/pthread_mutexattr_init

I think part of what's confusing me is that the multiple if statements are a strange way of writing the code, it didn't really occur to me that all of the if statements would be executed.

I suspect that the `pthread_mutex_destroy` and `pthread_mutex_init` on the native handle are technically undefined behavior though.  It probably won't cause issues now, but who knows what might change in the future?


Would a lockfree queue work, and/or changing the `std::mutex` to a `shared_mutex` work better than messing around with the internal mutex?  The original patch that was added in used effectively a `shared_mutex` with the lockfree queue(see: https://issues.apache.org/jira/browse/LOGCXX-500).  I think I removed everything that used `shared_mutex` to make it C++11 compatible, but there's still a compile test to see if it there(so we can easily switch between `std::shared_mutex` or `boost::shared_mutex`).

swebb2066: > Would a lockfree queue work
No, the writes by an appender must be serialized.

> or changing the std::mutex to a shared_mutex
No, to prevent priority inversion, the low priority thread that holds the lock must have its priority raised (temporarily), which is an operating system thing.

rm5248: My understanding of this is as follows:
1. you have multiple threads logging to the same appender that have different priorities
2. that appender is NOT an `AsyncAppender`, so there is a priority inversion when a low-priority thread has IO control

If that is correct, I now understand better what this is doing.

Is there a reason that `AsyncAppender` is not appropriate in your situation?  It seems like that would be an easier way to solve the problem, since all of the IO then happens in a different thread.

swebb2066: Using AsyncAppender could reduce the likelihood, but still result in a priority inversion. However, AsyncAppender would also increase context switching, which is not create on a real time controller.

A priority inversion problem is incredibily hard to verify and in my case only appears on deployed systems in the field. You cannot prove its absense with even extensive testing.

I believe it is an unfortunate accident of history the glibc does not do priority inheritance by default. I guess this is due to its relative recent discovery. Fortunately most applications do not have real time requirements. However the robotics community (which is a significant user of log4cxx - see ROS) really needs configurable logging and real time capabilities.

rm5248: Thanks for the explanation, that makes much more sense now.  I've never had to deal with a priority inversion before, so this has been a good learning experience.

One final thought: would it make more sense to create our own version of `std::mutex` that does the `pthread_mutexattr_setprotocol` in the constructor instead of effectively destroying and then re-initializing the underlying pthread mutex?

swebb2066:
Having looked deeper at the [Linux futex docs](https://man7.org/linux/man-pages/man2/futex.2.html) and [the glibc pthread implementation](https://github.com/lattera/glibc.git/nptl/pthread_mutex_lock.c), I now believe glibc does do priority inheritance locking by default.

It looks like it is not necessary to explicitly set the PRIO_INHERIT protocol on a mutex. `pthread_mutex_lock.c` only uses `FUTEX_LOCK_PI` and does not use `FUTEX_WAIT`.

This PR is no longer necessary

swebb2066:
Setting PTHREAD_PRIO_INHERIT does not change the behaviour of pthread_mutex_lock

PR #204, opened 2023-04-21
Allow message strings containing quote characters to be logged to a database

swebb2066:
The PR uses an option on the message specifier of the PatternLayout to duplicate any quote character in the message.

### DIFF
--- a/src/main/cpp/messagepatternconverter.cpp
+++ b/src/main/cpp/messagepatternconverter.cpp
@@ -20,12 +20,12 @@
 #include <log4cxx/spi/loggingevent.h>
 #include <log4cxx/spi/location/locationinfo.h>
 
-namespace log4cxx
-{
-namespace pattern
-{
+
+using namespace log4cxx;
+using namespace log4cxx::pattern;
 
 IMPLEMENT_LOG4CXX_OBJECT(MessagePatternConverter)
+
 /**
  * Formats the message of an logging event for a quoted context
   */
@@ -83,7 +83,4 @@ void MessagePatternConverter::format
 {
 	toAppendTo.append(event->getRenderedMessage());
 }
-
-} // namespace pattern
-} // namespace log4cxx
 
--- a/src/main/cpp/messagepatternconverter.cpp
+++ b/src/main/cpp/messagepatternconverter.cpp
@@ -20,31 +20,70 @@
 #include <log4cxx/spi/loggingevent.h>
 #include <log4cxx/spi/location/locationinfo.h>
 
-using namespace log4cxx;
-using namespace log4cxx::pattern;
-using namespace log4cxx::spi;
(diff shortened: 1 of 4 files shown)


rm5248:
Two things:
1. If this is to prevent SQL injection, shouldn't we use bound parameters instead?
2. It looks like the ABI broke; I suspect that this is because `IMPLEMENT_LOG4CXX_OBJECT(MessagePatternConverter)` is now inside of a namespace.

swebb2066:
1) This was inteneded as a bug fix, not to prevent a SQL an injection attack. Was it a vunerability?

2) Thanks for the ABI fix - I was wondering what it was.

rm5248:
My assumption was that the bug that you saw was that the SQL statement was not parsed correctly when you had a `'` character in the log message, since that delineates a string.

The proper way to insert stuff into the database is to use the [`SQLBindParameter`](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlbindparameter-function?view=sql-server-ver16) function in ODBC, so that there won't be any SQL injection and you don't have to worry about quoting the string.  So as it stands today I think we do have a SQL injection vulnerability.

I don't have a really good idea of how we should change the XML though, maybe something like:
```
<param name="sql" value="INSERT INTO [SomeDatabaseName].[SomeUserName].[SomeTableName] ([Thread],[LogName],[LogTime],[LogLevel],[FileName],[FileLine],[Message]) VALUES (?, ?, ?, ?, ?, ?, ?)" />
<param name="col1" value="%t"/>
<param name="col2" value="%c"/>
<param name="col3" value="%d{dd MMM yyyy HH:mm:ss.SSS}"/>
<param name="col4" value="%p"/>
<param name="col5" value="%f"/>
<param name="col6" value="%L"/>
<param name="col7" value="%m"/>
```

rm5248:
If we want to be fancy we could do something like what log4j2 does as well, and increment the column number whenever we see the param, so the XML would instead look like:
```
<param name="sql" value="INSERT INTO [SomeDatabaseName].[SomeUserName].[SomeTableName] ([Thread],[LogName],[LogTime],[LogLevel],[FileName],[FileLine],[Message]) VALUES (?, ?, ?, ?, ?, ?, ?)" />
<param name="ColumnMapping" value="%t"/>
<param name="ColumnMapping" value="%c"/>
<param name="ColumnMapping" value="%d{dd MMM yyyy HH:mm:ss.SSS}"/>
<param name="ColumnMapping" value="%p"/>
<param name="ColumnMapping" value="%f"/>
<param name="ColumnMapping" value="%L"/>
<param name="ColumnMapping" value="%m"/>
```

swebb2066:
I will start a new PR for the bound parameter change.

Does this vunerability need to be reported?

PR #205, opened 2023-04-22
Add support for using SQLPrepare and bound parameters to the ODBC appender

swebb2066:


swebb2066:
The initial implementation does not even work on Windows - The ParameterType argument of SQLBindParameter does not accept the value SQL_DEFAULT.

Not sure had to resolve these issue.

rm5248:
I have an idea of how to make it work, I will try it out later today or tomorrow.

rm5248:
I have the binding _somewhat_ working correctly, but not completely.  I think there's some mismatch between what I'm trying to do in C++ and what the C API will accept, e.g. the parameters being passed to ODBC are taking the address of temporaries and they're getting deleted before the SQL statement is executing fully.

swebb2066:
On Windows 10 (SQLExpress database) it now fails with - SQLGetDiagRecW reports `[Microsoft][ODBC SQL Server Driver]Invalid precision value`

swebb2066:
It seem the way to determine the parameter type is [SQLDescribeParam](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqldescribeparam-function)

However, according to [this helpful page](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/describing-parameters)is not widely supported.

Hence retaining the SQLExecDirect method capability using PatterLayout to modify the template SQL statement  is still required.

swebb2066:
The SQLBindParam now works on Windows bu executing the prepared statement results in the error `[Microsoft][ODBC SQL Server Driver]String data, right truncation`.

I will need to study the issue described [here ](https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/known-issues-in-this-version-of-the-driver) to work out what I am doing wrong.

rm5248:
thinking about this more, why don't we just use the [APR database functions](https://apr.apache.org/docs/apr-util/1.6/group___a_p_r___util___d_b_d.html) instead of trying to do our own?

rm5248:
created a quick test of using the APR functions in #206 - utilizing APR makes the code much much easier to use and work with and should support multiple databases.  I would recommend abandoning this PR in favor of the new one.

swebb2066:
This is now working to my satisfaction on Windows 10 with a SQLExpress back end.

PR #207, opened 2023-04-25
ODBCAppender support for logging MDC values

swebb2066:
Allows either a single value or all values (as JSON) of an event's mapped diagnostic context to be sent to a database field

### DIFF
--- a/src/main/include/log4cxx/patternlayout.h
+++ b/src/main/include/log4cxx/patternlayout.h
@@ -88,7 +88,7 @@ LOG4CXX_LIST_DEF(FormattingInfoList, log4cxx::pattern::FormattingInfoPtr);
  *  <tr>
  *      <td align="center"><strong>c</strong></td>
  *      <td>
- *          Used to output the logger of the logging event. The logger conversion specifier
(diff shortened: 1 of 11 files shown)


src/main/include/log4cxx/patternlayout.h
```diff
@@ -224,14 +224,22 @@ LOG4CXX_LIST_DEF(FormattingInfoList, log4cxx::pattern::FormattingInfoPtr);
...
+ *          the <strong>X</strong> is followed by a key placed between braces, as in <strong>%X{clientNumber}</strong>
+ *          where <code>clientNumber</code> is the key. In this case the value in the MDC corresponding to
  *          the key will be output.
  *          <p>See MDC class for more details.</p>
  *      </td>
  *  </tr>
  *  <tr>
+ *      <td align="center"><strong>J</strong></td>
+ *      <td>
+ *          Used to output JASON key/value pairs of all MDC (mapped diagnostic context)
```
rm5248, line 237: I've never seen it as JASON, only as JSON - is there some acronym that I'm missing?

swebb2066: OOPS

PR #210, opened 2023-04-29
Document how to send events to a fluent-bit server

swebb2066:
A fluent-bit server provides a [large list of log targets](https://docs.fluentbit.io/manual/pipeline/outputs).


src/main/include/log4cxx/appenderskeleton.h
```diff
@@ -146,8 +146,6 @@ class LOG4CXX_EXPORT AppenderSkeleton :
 		/**
 		Set the layout for this appender. Note that some appenders have
 		their own (fixed) layouts or do not use one. For example, the
```
rm5248, line 148: It looks like you forgot to delete the first part of the sentence

PR #211, opened 2023-05-01
Remove the layout for the ODBC appender, since we only insert into co…

rm5248:
…lumns


swebb2066:
What were you trying to achieve here?

PR #212, opened 2023-05-03
ODBCAppender should not use the message as an SQL statement

swebb2066:
This PR restores the 1.0 functionality (restores compatability).

Commit 0be583c caused the ODBC appender to use the message as the SQL statement.


swebb2066:
Made redundent by #213

PR #213, opened 2023-05-03
Column mappings are required to insert messages

rm5248:


swebb2066:
Also need to change the documentation in odbcappender.h

--- a/src/main/include/log4cxx/db/odbcappender.h
+++ b/src/main/include/log4cxx/db/odbcappender.h
@@ -56,16 +56,6 @@ either in the Log4cxx configuration file
 using the <b>sql</b> parameter element
 or programatically by calling the <code>setSql(String sql)</code> method.

-If no <b>ColumnMapping</b> element is provided in the configuration file
-the sql statement is assumed to be a PatternLayout layout.
-In this case all the conversion patterns in PatternLayout
-can be used inside of the statement. (see the test cases for examples)
-
-If the <b>sql</b> element is not provided
-and no <b>ColumnMapping</b> element is provided
-the attached a PatternLayout layout element
-is assumed to contain the sql statement.
-
 The following <b>param</b> elements are optional:
 - one of <b>DSN</b>, <b>URL</b>, <b>ConnectionString</b> -
   The <b>serverName</b> parameter value in the <a href="https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlconnect-function">SQLConnect</a> call.
@@ -138,10 +128,6 @@ An example configuration that writes to the data source named "LoggingDSN" is:
   <priority value ="INFO" />
   <appender-ref ref="ASYNC" />
 </root>
-<appender name="PatternAppender" class="ODBCAppender">
- <param name="DSN" value="LoggingDSN"/>
- <param name="sql" value="INSERT INTO [ApplicationLogs].[dbo].[UnitTestLog] ([Thread],[LogName],[LogTime],[LogLevel],[FileName],[FileLine],[Message],[MappedContext]) VALUES ('%t', '%c','%d{yyyy-MM-dd HH:mm:ss.SSSSSS}','%p','%f','%L','%m{'}','%J{'}')" />
-</appender>
 </log4j:configuration>
 ~~~

swebb2066 (approved):

PR #216, opened 2023-06-16
Fix deallocation bug in rolling policy base private.

landwehrj:
This fixes a deallocation bug from the parent pointer.

Rolling policy base private needs a virtual destructor because the base pointer to this class is being used to destruct things.

An example of this is in fixedwindowrollingpolicy.cpp:16 where the FixedWindowRollingPolictyPrivate is being passed to RollingPolicyBase() which stores in m_priv which is RollingPolicyBasePrivate.

When this pointer deallocates it will have the wrong size.

To reproduce:
1) Compile log4cxx in debug. (cmake ../ -DCMAKE_BUILD_TYPE=DEBUG)
2) Compile bug.cpp with addresssantizer (g++ -g -fsanitize=address  bug.cpp)
3) Run bug.cpp with audit.properties

These files are in the test.zip.

You should see something similar to result.

[test.zip](https://github.com/apache/logging-log4cxx/files/11776735/test.zip)

--------------------

As an aside to this bug, if you remove line 24 of bug.cpp it will cause a segfault because it is using a file watcher which uses a static variable that gets deallocated on main exit and other threads are using it.

Specifically: loglog.cpp:59 will become deallocated.
And the filewatchdog.cpp:126 will call getInstance().

### DIFF
--- a/src/main/include/log4cxx/private/rollingpolicybase_priv.h
+++ b/src/main/include/log4cxx/private/rollingpolicybase_priv.h
@@ -28,6 +28,8 @@ namespace log4cxx
 namespace rolling{
 
 struct RollingPolicyBase::RollingPolicyBasePrivate {
+    virtual ~RollingPolicyBasePrivate() = default;
+
     /**
      * File name pattern converters.
      */


swebb2066:
I think the warning reported by AddressSanitizer (new-delete-type-mismatch) does not generally cause a problem - The full size memory block will be free()ed and returned to the heap. 

The [sized delete](https://isocpp.org/files/papers/n3778.html) introduced in C++14 enables asan_new_delete.cpp implementation to issue a warning as GCC seems to now use it, but the unsized delete is equivalent to the sized delete.

swebb2066:
To avoid clutering the base class by introducing a virutal destructor, we could use the following change:
```
--- a/src/main/cpp/fixedwindowrollingpolicy.cpp
+++ b/src/main/cpp/fixedwindowrollingpolicy.cpp
@@ -57,7 +57,10 @@ FixedWindowRollingPolicy::FixedWindowRollingPolicy() :
 {
 }

-FixedWindowRollingPolicy::~FixedWindowRollingPolicy(){}
+FixedWindowRollingPolicy::~FixedWindowRollingPolicy()
+{
+    delete static_cast<FixedWindowRollingPolicyPrivate*>(m_priv.release());
+}
```
To address `new-delete-type-mismatch` completely the following files would also need to be changed:
```
andfilter.cpp
aprdatagramsocket.cpp
aprserversocket.cpp
aprsocket.cpp
asyncappender.cpp
colorstartpatternconverter.cpp
consoleappender.cpp
datepatternconverter.cpp
dbappender.cpp
fileappender.cpp
filerenameaction.cpp
fixedwindowrollingpolicy.cpp
gzcompressaction.cpp
levelmatchfilter.cpp
levelrangefilter.cpp
literalpatternconverter.cpp
locationinfofilter.cpp
loggermatchfilter.cpp
mapfilter.cpp
multiprocessrollingfileappender.cpp
namepatternconverter.cpp
nteventlogappender.cpp
odbcappender.cpp
propertiespatternconverter.cpp
rollingfileappender.cpp
smtpappender.cpp
socketappenderskeleton.cpp
stringmatchfilter.cpp
syslogappender.cpp
telnetappender.cpp
writerappender.cpp
xmlsocketappender.cpp
```

rm5248:
> I think the warning reported by AddressSanitizer (new-delete-type-mismatch) does not generally cause a problem - The full size memory block will be free()ed and returned to the heap.

I'm not sure if it actually is - I had fixed a few ASAN errors a while ago by adding [virtual destructors to the base class](https://github.com/apache/logging-log4cxx/commit/a46055a65b29e828d5bf85c2595031f78672b429), because it was leaking memory(a string was being allocated in the parent class, but the parent destructor was never called so the string was never deleted).

Adding the virtual destructor is a simple one-line fix instead of modifying ~20 files.

landwehrj:
Thank you.

PR #218, opened 2023-06-19
jsonlayout, register the class in class.cpp

petitlapin:
Hi,
When linking statically with log4cxx, I got the issue:
```
log4cxx: Count not instantiate class [org.apache.log4j.JSONLayout]
log4cxx: Class not found: org.apache.log4j.JSONLayout
```

Registering the class solves the issue. I could not pinpoint on why the class was not present in the executable.

### DIFF
--- a/src/main/cpp/class.cpp
+++ b/src/main/cpp/class.cpp
@@ -46,6 +46,7 @@
 #include <log4cxx/net/xmlsocketappender.h>
 #include <log4cxx/layout.h>
 #include <log4cxx/patternlayout.h>
+#include <log4cxx/jsonlayout.h>
 #include <log4cxx/htmllayout.h>
 #include <log4cxx/simplelayout.h>
 #include <log4cxx/xml/xmllayout.h>
@@ -172,6 +173,7 @@ void Class::registerClasses()
 	log4cxx::nt::OutputDebugStringAppender::registerClass();
 #endif
 	SMTPAppender::registerClass();
+	JSONLayout::registerClass();
 	HTMLLayout::registerClass();
 	PatternLayout::registerClass();
 	SimpleLayout::registerClass();


swebb2066:
I believe this problem is somewhat pervasive (it is a pattern used in other places) and requires a more extensive change. 

The `IMPLEMENT_LOG4CXX_OBJECT` macro which creates a statically initialised variable (e.g. JSONLayoutRegistration) using the code:
```
namespace log4cxx { namespace classes { \
	const ::log4cxx::helpers::ClassRegistration& object##Registration = object::registerClass(); \
	} }
```
was placed in the jasonlayout.cpp translation unit instead of loader.cpp. The jasonlayout.cpp translation unit is not included in log4cxx.so as no other log4cxx translation unit references JSONLayout.

I think we should move all instances of `IMPLEMENT_LOG4CXX_OBJECT` to loader.cpp.

Is there another way?

rm5248:
> I think we should move all instances of `IMPLEMENT_LOG4CXX_OBJECT` to loader.cpp.

I'm not in favor of that, since if/when we move things to plugins we don't want the plugins referenced in the log4cxx.so file.  Keeping the implementation contained to the .cpp files seems cleanest to me.

> Is there another way?

This is the simplest way with the minimal number of changes.  

We could do something like adding `attribute((constructor))` to the `IMPLEMENT_LOG4CXX_OBJECT` macro that would do the same thing(call `registerClass` when the .so is loaded), but that is GCC/Clang specific.

swebb2066:
It seems that GCC and MSVC do include JSONLayout in the DSO. The (normally disabled) `test_fluent_bit()` test in `xmlsocketappendertestcase` uses JSONLayout in the configuration file. I successfully tested that on Windows and Linux.

@petitlapin what compiler are you using?

petitlapin:
> @petitlapin what compiler are you using? Are you configuring log4cxx in a static initializer (e.g. in a getLogger() wrapper)?

I've tested with ICC 2021.4 and ICX 2023.0 (which is clang16.0 with some improvement) in a ubi8 docker image. We compile apr, aprutil and log4cxx statically. The log is used in a static initializer as you say.

I'll try to get a minimal example to reproduce it.

swebb2066:
The problem will be that no code references JSONLayout (either in log4cxx or your code) so the linker would not have included the log4cxx.a jsonlayout.o translation unit in your executable image.

The problem is not as extensive as I feared. Log4cxx code includes references to most classes named in `IMPLEMENT_LOG4CXX_OBJECT` macros. The exceptions are some appenders, layouts and filters.

Issue #233, opened 2023-07-17
Ability to remove a HierarchyEventListener?


TroudhuK:
Hierarchy::addHierarchyEventListener allows to register a listener to intercept the appenders additions to a logger.
But there is no way to remove a listener, for example because its lifetime isn't the same as the logger (when we do it to Logger::getRootLogger()->getLoggerRepository()). True, or do I miss something?

If true, I'd like to have a Hierarchy::removeHierarchyEventListener.

rm5248:
It looks like this was never added, since it is not part of the Log4j1 API.  It seems reasonable to add though.

rm5248:
fixed in #234

PR #234, opened 2023-07-17
Add a way to remove the HierarchyEventListener

rm5248:
See #233


rm5248:
Note: ABI failure is due to adding new virtual function(see: https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B).  The important note:

> You cannot:
> Add new virtual functions for any reason, even to leaf classes, _if the class is intended to remain binary compatible on Windows._

So I think we will just have to add this to only the `Hierarchy` class for now, ignoring the parent `LoggerRepository`.  Whenever we bump the ABI we can fix this.

PR #238, opened 2023-07-23
Improve the structure, clarity and correctness of documentation

swebb2066:
I believe the sub-sections under "Loggers, Appenders and Layouts" would be better placed under "Usage".


swebb2066:
The [generated web site](https://logging.staged.apache.org/log4cxx/latest_stable/usage-overview.html) is available for review.

rm5248:
Minor notes:

- The 'layouts' section that is currently in the 'Loggers, appenders and layouts' section may want to be instead in the 'Configuration samples' page.  Or we should link to that from the 'Configuration Samples' page since it talks about updating the pattern layout for one of the XML files.
- On the filters page, it may make sense to add in the anchors (`{#labeling-log-output}`) so the TOC won't have doxygen auto-generated links.  Also apparently 'labeling' can have one L or two, I did not realize that.
- The 'examples program' page is currently called 'coding'. Maybe that should be renamed to something like 'quick-start' and be the first page under 'usage'?

Another thing TODO(may be good for separate PR): on the 'releases' page, we should have a link to the documentation for that particular release(help prevent link rot, so you can get documentation for a particular release).  e.g. 1.0.0 doc is here: https://logging.apache.org/log4cxx/1.0.0/

swebb2066:
The [updated web site](https://logging.staged.apache.org/log4cxx/latest_stable/usage-overview.html) is available for review.

Note: The left navigation panel cache does not seem to update on F5 - I had to clear the browser cache.

PR #241, opened 2023-07-27
Remove CFString support

swebb2066:
The original commit by Curt Arnold was 2007 (b0ce6ef31). It has not been compiled since 0.10.0.


rm5248:
Maybe it worked originally?  The API may have changed since 2007.

I don't feel comfortable removing a feature that should work without a major bump in the ABI though, since if we remove the CFString methods that is an ABI-breaking change(at least on OSX).

swebb2066:
OK, it stays

swebb2066:
I cannot see how to make CFString compile on OSX.

Any suggestions?

swebb2066:
It seems it is std::basic_string<UniChar> that is not supported.

PR #251, opened 2023-08-12
Make utf-8 the default character encoding

swebb2066:
The current default "locale" is not useful as it actually reverts to "US-ASCII" which writes a '?' for any non-ascii character.


rm5248:
If you're getting ? characters, you may just need to set your locale(see FAQ: https://logging.apache.org/log4cxx/latest_stable/faq.html#unicode_supported)

I'm not sure if this changes that in any bad way.

swebb2066:
Unfortunately, the LocaleCharsetEncoder implementation has never worked. The names reported by `apr_os_locale_encoding` do not correspond to the names recognised by `CharsetEncoder::getEncoder()`

swebb2066:
With the LocaleCharsetEncoder now fixed, the change to UTF-8 as the default would only be justified if UTF-8 is a better default for log files, which I suspect is the case.

@rm5248 do you know what log file encode aggregation services usually expect?

swebb2066:
While the encoder can be specified in the appender configuration, the examples in the documentation do not include tht option.

Sending a JSONLayout or an XMLLayout to a file or socket in anything other than UTF-8 would not be good, and UTF-8 is an expected default for .txt files. 

Hence I conclude that changing the default encoding to UTF-8 is important.

PR #254, opened 2023-08-13
Simplify CMake build configuration

swebb2066:
This PR make the build behaviour align with the documented behaviour.

> If C++17 is not available, a search for Boost will be taken and those libraries will be used instead.

Also, remove mention of unused configuration variables.


rm5248:
Note: the reason that the `PREFER_BOOST` flag exists is to fix this issue: https://issues.apache.org/jira/browse/LOGCXX-551

It looks like this is fine though, the only things removed are C++11 features that must already exist.

PR #264, opened 2023-08-19
Update the change report

swebb2066:
This PR adds entries to the change report while my memory is fresh

### DIFF
--- a/src/site/markdown/change-report-gh.md
+++ b/src/site/markdown/change-report-gh.md
@@ -49,10 +49,15 @@ Changelog {#changelog}
 
 The following issues have been addressed:
 
-* The `locale` external character encoding option reverting to US-ASCII
-* Unnecessary string copying when sending a logging event to an appender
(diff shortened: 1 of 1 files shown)


src/site/markdown/change-report-gh.md
```diff
@@ -44,6 +45,17 @@ Changelog {#changelog}
...
+
+The following issues have been addressed:
+
+* The `locale` external character encoding option reverting to US-ASCII
+* Unnecessary string copying when sending a logging event to an appender
+* `Qt support/integration` not supporting a QString value in a LOG4CXX_XXXXX macro
+* A segfault during process exit when LogManager::shutdown is not called.
+
+Note: the `LOG4CXX_CHARSET` cmake option (external character encoding) default value has changed to `utf-8`
+
```
rm5248, line 58: should also add:
* hexdump utility added to log bytes
* Ability to remove HierarchyEventListener(#233)

For the Qt support, I would phrase it more that "LOG4CXX_XXXX macros now support logging of QStrings directly" to indicate that this is a new feature.

PR #267, opened 2023-09-12
Revert to console output when configuration file is missing in examples

swebb2066:
Sending logging events to the console seems to be more useful than seeing the message

> Please initialize the log4cxx system properly.

on the console.

### DIFF
--- a/src/examples/cpp/com/foo/config-qt.cpp
+++ b/src/examples/cpp/com/foo/config-qt.cpp
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 #include "config-qt.h"
+#include <log4cxx/basicconfigurator.h>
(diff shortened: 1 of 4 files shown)


rm5248:
It looks like that message will still be printed out, is that correct?  I want to make sure that it will still indicate to the user somehow that configuration was not successful, since otherwise it could look like it succeeded when it actually failed.

swebb2066:
This PR only changes the documented examples (i.e. it is in the user's domain). The return value of *::configure tells the user that configuration was not successful. They could handle the failure differently. An important result of this PR is to show that the failure to configure is available in the return value of *::configure.

A separate PR could be raised to configure console logging after that message is printed out if you think it appropriate.

PR #277, opened 2023-10-29
Custom namespace support

vopl:
The possibility of placing the logger functionality in a custom namespace allows you to use several different versions of the logger within the same process, which in turn allows you to create complex products that include binary code from different vendors, regardless of each other using a convenient version of the logger.
The motivation for this approach is exactly the same as in Qt: https://wiki.qt.io/Qt_In_Namespace


rm5248:
It looks like some symbols were removed somehow, which is why the ABI checks fail.  Perhaps the namespace was not set correctly?

compat report can be downloaded from here: https://github.com/apache/logging-log4cxx/actions/runs/6683446793

vopl:
These are:
- extern "C" void tlsDestruct(void* ptr);
should not be exported
- class DiscardSummary with all methods
should not be exported
- uint32_t libraryVersion()
should be exported within the log4cxx namespace, but exported globally

vopl:
Ahh.. wrongly closed by missclick, reopen

swebb2066:
How does this affect the doxygen generated site?

vopl:
A doxygen generated site works fine

vopl:
The [ubuntu-22.04-clang++-build-and-test](https://github.com/apache/logging-log4cxx/actions/runs/6684814433/job/18163287628?pr=277#logs) action failed due to a broken toolchain combination of clang14 plus libc++ from gcc-13. This is not a problem of this pull-request.

rm5248:
Is this to fix a problem that you have encountered before?  I'm wondering because having the ABI stable(with version 1.0.0 and above) should help fix that problem.

I generated the site and it looks fine to me, the namespace seems to be generated fine.

Could you add some documentation on how to use this feature?  And ideally when you would use this feature.

vopl:
> Is this to fix a problem that you have encountered before? I'm wondering because having the ABI stable(with version 1.0.0 and above) should help fix that problem.

All commits from this pull-request are targeted to а possibility, declared at first message here: https://github.com/apache/logging-log4cxx/pull/277#issue-1966976922

ABI keeps stable and not changed by default, but totally changed if `LOGCXX_NS=log4cxx' CMake variable changes, and this is a wanted goal - allow to use more then one ABIs from different loggers in moment without any conflicts.

> Could you add some documentation on how to use this feature? And ideally when you would use this feature.
My documentor-skills are very bad :( The central changer here is LOGCXX_NS variable. If it has a value 'log4cxx' (default), then all happens as before, no changes. If LOGCXX_NS=somecustomname, then all log4cxx entities are placed in that 'somecustomname' namespace instead of 'log4cxx'. This allows to use two and more logger instances simultaneously, for example, 'some old plugin X' uses log4cxx instance from local system, central application uses own log4cxx with LOGCXX_NS=xyz, other plugin uses thrid log4cxx instance with LOGCXX_NS=qwe. All components are not constrained by each other an now and can made an unrestricted (by other parts through log4cxx) development progress.

swebb2066:
Would it be helpful to have a new documentation usage section about using Log4cxx in components?

If we consider this PR as a part of the solution to using Log4cxx in components, the usage documentation could include a couple of senarios and specific examples. I suggest such a section should address how common package managers could then handle Log4cxx.

Also, it should provide some Log4cxx configuration patterns with examples.

PR #281, opened 2023-11-05
Flush at exit

vopl:
Flushing all buffered data during the static deinitialization phase.

### DIFF
--- a/src/examples/cpp/custom-appender.cpp
+++ b/src/examples/cpp/custom-appender.cpp
@@ -38,8 +38,6 @@ public:
 		return false;
 	}
 
-	void flushBuffers() override{}
-
 	void append(const spi::LoggingEventPtr& event, log4cxx::helpers::Pool& p) override {
 		// This gets called whenever there is a valid event for our appender.
 	}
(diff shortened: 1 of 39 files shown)


rm5248:
Shouldn't `LogManager::shutdown` do the proper flushing of data?  Or does that not work under certain situations?

vopl:
> Shouldn't `LogManager::shutdown` do the proper flushing of data? Or does that not work under certain situations?

`LogManager::shutdown` terminates logger functionality, but this case about events-at-exit, in which all logger functionality should continue to work

swebb2066:
Perhaps you also could include SocketOutputStream and ODBCAppender?

vopl:
I can't support SocketOutputStream, it doesn't have enough internal infrastructure for flushing, in particular, there is no  apr-Pool instance, which is placed in more derived types. (It's not actually used for flushing, but I don't able to make a second layer of stubs)

swebb2066:
> I can't support SocketOutputStream
OK, but as XMLSocketAppender calls flush() in each append(), I suggest the change [XMLSocketAppender](https://github.com/apache/logging-log4cxx/pull/281/commits/876763f1a06781a19dbca5f1a767f838c932a078) is not required.

PR #282, opened 2023-11-05
Ide friendly

vopl:
Indexing the header files in a fictitious target to work more conveniently with them in the IDE (navigation, refactoring, etc)

### DIFF
--- a/src/main/cpp-qt/CMakeLists.txt
+++ b/src/main/cpp-qt/CMakeLists.txt
@@ -15,30 +15,28 @@
 # limitations under the License.
 #
 
-option(LOG4CXX_QT_SUPPORT "Qt support/integration" OFF)
-
 # Log4cxx Qt support, if enabled
 if(LOG4CXX_QT_SUPPORT)
     find_package(Qt5 COMPONENTS Core REQUIRED)
 
     add_library(log4cxx-qt)
     if(BUILD_SHARED_LIBS)
-    target_compile_definitions(log4cxx-qt PRIVATE LOG4CXX)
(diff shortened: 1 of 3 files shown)


rm5248:
note: I need to take a closer look at this, since I see there is a random file that is generated I want to make sure it's not going to mess anything up with symbols in the library.  Otherwise I would merge it

vopl:
All headers are available in IDE, so, IDE can does many useful things: text search, code completion, navigation between c++ entities, analyze sources with clang tools, ...
image

> note: I need to take a closer look at this, since I see there is a random file that is generated I want to make sure it's not going to mess anything up with symbols in the library. Otherwise I would merge it

One a fake source file `${CMAKE_CURRENT_BINARY_DIR}/empty-lib-stub.cpp`, I have added to include-target, is a minimal c++ translation unit with one an entity, named `someUnused`. It is required for some toolchains (on osx for example), which cannot works with empty archives. This entity has an internal linkage and does not used from any other, so, no symbol exported in result binary. Also, please note, include-targets do not link with general targets, so, no include-content is placed to them.

rm5248:
> One a fake source file `${CMAKE_CURRENT_BINARY_DIR}/empty-lib-stub.cpp`, I have added to include-target, is a minimal c++ translation unit with one an entity, named `someUnused`. It is required for some toolchains (on osx for example), which cannot works with empty archives. This entity has an internal linkage and does not used from any other, so, no symbol exported in result binary. Also, please note, include-targets do not link with general targets, so, no include-content is placed to them.

Ah, it's a separate target, I didn't see that at first.  Looks good to me; I'll merge it in tomorrow morning if there are no objections from anybody.

PR #288, opened 2023-11-10
fix reuse of a per thread cached stream

vopl:
This fix repetitive usage of per thread cached ostringstream in StringOrStream-CharMessageBuffer-MessageBuffer. 
The problem occurs when using output manipulators ([standart](https://en.cppreference.com/w/cpp/io/manip) or custom) that change the state of the stream in aspect of formatting output values. Example:
```
int num = 220;
LOG4CXX_INFO(logger, "some number in hex: " << std::hex << num);
LOG4CXX_INFO(logger, "wrongly hex number base here: " << num);
```

I am not sure that the proposed solution is optimal, perhaps it is worth simply abandoning caching of the ostringstream value.

### DIFF
--- a/src/main/cpp/messagebuffer.cpp
+++ b/src/main/cpp/messagebuffer.cpp
@@ -54,9 +54,21 @@ struct StringOrStream
 		if (!this->stream)
 		{
 #if LOG4CXX_HAS_THREAD_LOCAL
-			thread_local static std::basic_ostringstream<T> sStream;
-			this->stream = &sStream;
+			thread_local static struct
+			{
+				std::basic_ostringstream<T> sStream;
+				std::locale locale = sStream.getloc();
+				std::streamsize precision = sStream.precision();
+				std::ios_base::fmtflags flags = sStream.flags();
+				T fill = sStream.fill();
+			} streamWithInitialState;
+
+			this->stream = &streamWithInitialState.sStream;
 			this->stream->clear();
+			this->stream->imbue(streamWithInitialState.locale);
+			this->stream->precision(streamWithInitialState.precision);
+			this->stream->setf(streamWithInitialState.flags, ~streamWithInitialState.flags);
+			this->stream->fill(streamWithInitialState.fill);
 #else
 			this->stream = new std::basic_ostringstream<T>();
 #endif
--- a/src/main/cpp/messagebuffer.cpp
+++ b/src/main/cpp/messagebuffer.cpp
@@ -57,19 +57,12 @@ struct StringOrStream
 			thread_local static struct
 			{
 				std::basic_ostringstream<T> sStream;
-				std::locale locale = sStream.getloc();
-				std::streamsize precision = sStream.precision();
-				std::ios_base::fmtflags flags = sStream.flags();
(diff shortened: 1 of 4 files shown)


swebb2066:
The cached std::ostringstream significantly reduces the cost of logging when using `std::ostream& operator<<(std::ostream, ...)` overloads.

I suggest you check the before and after output of the Google benchmark test (built by defining BUILD_BENCHMARK_CHECKS=on )

vopl:
I have no performance requirements at the moment, I cannot adequately evaluate the benchmark results. Offhand, the proposed solution does not significantly reduce performance.

rm5248:
There are a few ideas on fixing this here: https://stackoverflow.com/questions/2273330/restore-the-state-of-stdcout-after-manipulating-it

It looks like what is proposed here does it correctly.  We should add a test to ensure that it works fine.

swebb2066:
Here is the relevant benchmark line before this PR:
`Logging int value with std::ostream                                   697 ns          696 ns      1017033`
and with commit [638d49d](https://github.com/apache/logging-log4cxx/pull/288/commits/638d49dd7daed7c02816895a60179409deb2c886):
`Logging int value with std::ostream                                   757 ns          757 ns       957090`
and after the change using `copyfmt`:
`Logging int value with std::ostream                                   828 ns          828 ns       828736`

vopl:
@swebb2066 can You please show an benchmark for disabled ostringstream cache?

rm5248:
I added a test to the branch to make sure that it works fine.  If you run the new test with current master, you will see that it does fail.

swebb2066:
> @swebb2066 can You please show an benchmark for disabled ostringstream cache?

A disabled logger (e.g. when Logger::isDebugEnabledFor() if false) is less than 1 ns on Ubuntu from memory. (I broke my Ubuntu trying to fix its wifi) On Windows using a dll (not inlined) it is abot 4 ns.

vopl:
> A disabled logger (e.g. when Logger::isDebugEnabledFor() if false) is less than 1 ns on Ubuntu from memory. (I broke my Ubuntu trying to fix its wifi) On Windows using a dll (not inlined) it is about 4 ns.

I mean a case in which the ostringstream instance is not cached by thread_local static, but is built again for each new instance of Message buffer.

stephen-webb:
>  I mean a case in which the ostringstream instance is not cached by thread_local static, but is built again for each new instance of Message buffer

`Logging int value with std::ostream                                   915 ns          915 ns       751322`

stephen-webb:
I suggest the following:
`                       const static std::basic_ostringstream<T> initialState;
                        thread_local static std::basic_ostringstream<T> sStream;
                        this->stream = &sStream;
                        this->stream->clear();
                        this->stream->precision(initialState.precision());
                        this->stream->setf(initialState.flags(), ~initialState.flags());
                        this->stream->fill(initialState.fill());
 which has this performance:
`Logging int value with std::ostream                                   684 ns          684 ns       982152`

vopl:
I also added `width`, it is small and will not introduce large slowdowns.

PR #291, opened 2023-11-16
Optionally use std::format in LOG4CXX_XXXX_FMT macros instead of fmt::format

swebb2066:
This PR allows the following usage options:

1. Add `-DLOG4CXX_FORMAT_NAMESPACE=std` to the cmake configure step when building Log4cxx
2. Add `#define LOG4CXX_FORMAT_NS std` before `#include <log4cxx/logger.h>`


rm5248:
Would it make sense to be able to switch this at compile time for the user?  We can turn off location information by defining the preprocessor argument `LOG4CXX_DISABLE_LOCATION_INFO`, we could do something similar here.  That can also let the user select between the two if needed, since fmt has more comprehensive options than std::format.

At least part of this could be done automatically by checking for `__cpp_lib_format`

swebb2066:
> Would it make sense to be able to switch this at compile time for the user?

Yes, which is done by adding `#define LOG4CXX_FORMAT_NS std` before `#include <log4cxx/logger.h>`

rm5248:
> > Would it make sense to be able to switch this at compile time for the user?
> 
> Yes, which is done by adding `#define LOG4CXX_FORMAT_NS std` before `#include <log4cxx/logger.h>` or `LOG4CXX_FORMAT_NS=std` as a preprocessor argument

It doesn't look like it checks for that though, I think you would get a warning/error about multiple defines.  I would assume it actually needs to be this:

```
#define LOG4CXX_USING_STD_FORMAT @LOG4CXX_USE_STANDARD_FORMAT@
#if !defined(LOG4CXX_FORMAT_NS) && LOG4CXX_USING_STD_FORMAT
#define LOG4CXX_FORMAT_NS std
#elif !defined(LOG4CXX_FORMAT_NS)
#define LOG4CXX_FORMAT_NS fmt
#endif
```

This information should probably also be added to the `macros influencing log4cxx` page.

swebb2066:
[Updated website](https://logging-log4cxx.staged.apache.org/log4cxx/latest_stable/) is available for review

PR #299, opened 2023-11-21
MDC and NDC availablity no longer depend on APR thread support

swebb2066:


src/main/cpp/threadspecificdata.cpp
```diff
@@ -60,13 +60,16 @@ ThreadSpecificData& ThreadSpecificData::getDataNoThreads()
 
 ThreadSpecificData* ThreadSpecificData::getCurrentData()
 {
-#if APR_HAS_THREADS
 	void* pData = NULL;
+#if APR_HAS_THREADS
```
rm5248, line 64: should these `#if` be reversed, such that if it has `thread_local` that takes precedence over APR?

swebb2066:
> should these #if be reversed, such that if it has thread_local that takes precedence over APR?

I have not yet sure that using thread_local instead of APR would make some some subtle behaviour change.

PR #301, opened 2023-11-27
Provide temporary logging level change support using RAII

swebb2066:
This PR add a Log4cxx extension to support complex algorithm development. I found this necessary in many cases to diagnose behaviour when process large data sets.

An example of my usage in a test is:
```
        log4cxx::ScopedLevelChange ctx2("ConfigurationTrajectoryCost", m_log);
        ProblemType::DurationVector solutionDurations = func.GetStepDurations(solver.Solution());
        LOG4CXX_DEBUG(m_log, "testAlternatingSpeeds:"
            << " solutionDuration " << solutionDurations.Sum()
            << ": " << solutionDurations
            );
        BOOST_CHECK_LE(solutionDurations.Sum(), initialGuessDurations.Sum());

```

### DIFF
--- a/src/main/include/log4cxx/levelchange.h
+++ b/src/main/include/log4cxx/levelchange.h
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


rm5248:
Depending on your use-case, what I have done before is to have each class have a logger(java-style) but then also define some other top-level loggers for certain long-term operations.

Another alternative here would be to add a filter (MapFilter?) that would log messages when a certain key is in the MDC.

swebb2066:
A use-case relevant to this PR is debugging computationally heavy processing. Activating DEBUG logging results in massive log files. Having identified the problematic function, it is useful to reduce the flood of output during further analysis.

> Long-term we should maybe think about implementing markers like log4j2 does: https://logging.apache.org/log4j/2.x/manual/markers.html

This requires changing the class under analysis, which is also valid approach, but more invasive.

PR #310, opened 2023-12-09
A build option to use the std::filesystem::path in place of log4cxx::File

swebb2066:
This PR is another experimental step towards removing APR as a dependency.

The principal remaining log4cxx::File function to replace is apr_file_open.


rm5248:
The ABI failures are a false positive, it's complaining about the private data structure size changing so that shouldn't cause a problem.

At some point we will need to break the ABI to remove the pool option, would we delete the File class at that point too in order to move entirely over to `std::filesystem`?

swebb2066:
> At some point we will need to break the ABI to remove the pool option, would we delete the File class at that point too in order to move entirely over to std::filesystem?

A log4cxx without APR does not now seem to me worth attempting. Without a socket appender,, XML config, zip/gzip log compression, dbappender, multiprocessrollingfileappender, etc it would be a different product. I think that product is [log4cplus](https://github.com/log4cplus/log4cplus).

I am inclined to abandom this PR now.

PR #316, opened 2023-12-19
Improve documentation of macros by labelling example usage

swebb2066:
The [updated web site](https://logging.staged.apache.org/log4cxx/latest_stable/usage-overview.html) is available for review.

### DIFF
--- a/src/main/include/log4cxx/logger.h
+++ b/src/main/include/log4cxx/logger.h
@@ -2056,19 +2056,21 @@ Add a new logging event containing \c message to attached appender(s) if this lo
 /**
 Add a new logging event containing \c message to attached appender(s) if \c logger is enabled for <code>DEBUG</code> events.
 
-@param logger the logger that has the enabled status.
-@param message a valid r-value expression of an <code>operator<<(std::ostream&. ...)</code> overload.
-
+\usage
 ~~~{.cpp}
 LOG4CXX_DEBUG(m_log, "AddMesh:"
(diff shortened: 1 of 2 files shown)


src/main/include/log4cxx/logger.h
```diff
@@ -2056,19 +2056,21 @@ Add a new logging event containing \c message to attached appender(s) if this lo
...
 Add a new logging event containing \c message to attached appender(s) if \c logger is enabled for <code>DEBUG</code> events.
 
-@param logger the logger that has the enabled status.
-@param message a valid r-value expression of an <code>operator<<(std::ostream&. ...)</code> overload.
-
+\usage
 ~~~{.cpp}
 LOG4CXX_DEBUG(m_log, "AddMesh:"
 	<< " name " << meshName
 	<< " type 0x" << std:: hex << traits.Type
```
rm5248, line 2063: It's not something that you modified here, but won't the `std::hex` here also affect the later calls of `getCenter()` and `getHalfSize()`? I think the two examples(LOG4CXX_DEBUG and LOG4CXX_DEBUG_FMT) are not equivalent in that case.

swebb2066: > the later calls of getCenter() and getHalfSize()

These actually return 3d float vectors which are handled by a `operator<<(std::ostream& ` overload. Not obvious I know, but I wanted to use a real example.

PR #325, opened 2024-01-07
Remove unnecessary delays in unit tests

swebb2066:
The delay in VectorAppender does not seem to be currently required.


rm5248:
Do we even need _any_ delays?  I think with the clock mocking that I did a while ago we don't need any delays, we just advance our fake clock the correct number of milliseconds.  It does it in the timebasedrollingtest only at the moment.

swebb2066:
> Do we even need any delays?

The reason for the delay is known only to Michael CATANZARITI who created vertorappender.cpp in 2003 (see commit fc5053e14).

On closer inspection I think the delay was for asyncappendertestcase test3 to use blocking to ensure the async thread does not remove all the logging events until the appender is closed.

I will update the code using this new hypothesis

PR #335, opened 2024-01-22
Improve AsyncAppender throughput

swebb2066:
This PR aims to reduce synchronization contention when logging through AsyncAppender.

Relevant Windows benchmarks improve from:

 | Benchmark | Time | CPU | Iterations | 
 | --------- | ----- | ----- | --------- |
 | Async, int value using MessageBuffer, pattern: %m%n | 1594 ns | 928 ns | 640000 | 
 | Async, int value using MessageBuffer, pattern: %m%n/threads:4 | 687 ns | 1445 ns | 400000 | 

to:

 | Benchmark | Time | CPU | Iterations | 
 | -------- | ---- | ----- | ------------------- | 
 | Async, int value using MessageBuffer, pattern: %m%n | 1654 ns | 879 ns | 1120000 | 
 | Async, int value using MessageBuffer, pattern: %m%n/threads:4 | 671 ns | 1484 ns | 400000 |

### DIFF
--- a/src/main/cpp/asyncappender.cpp
+++ b/src/main/cpp/asyncappender.cpp
@@ -551,7 +551,7 @@ void AsyncAppender::dispatch()
 			priv->bufferNotEmpty.wait(lock, [this, &eventList]() -> bool
 				{
 					eventList = priv->eventList.pop_all_reverse();
-					return !eventList || priv->closed;
+					return eventList || priv->closed;
 				}
 			);
 			isActive = !priv->closed;
--- a/src/main/cpp/asyncappender.cpp
+++ b/src/main/cpp/asyncappender.cpp
@@ -555,7 +555,6 @@ void AsyncAppender::dispatch()
 					return eventList || priv->closed;
 				}
 			);
-			isActive = !priv->closed;
 		}
 		priv->approxListSize = 0;
 		priv->bufferNotFull.notify_all();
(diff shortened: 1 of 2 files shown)


swebb2066:
Relevant Ubuntu gcc 9 benchmarks improve from:

 | Benchmark | Time | CPU | Iterations | 
 | -------- | ------ | ---- | ----------- | 
 | Async, int value using MessageBuffer, pattern: %m%n | 605 ns | 605 ns | 1171481 | 
 | Async, int value using MessageBuffer, pattern: %m%n/threads:16 | 997 ns | 8010 ns | 89648 | 

to:

 | Benchmark | Time | CPU | Iterations | 
 | -------- | ---- | --- | ------------ | 
 | Async, int value using MessageBuffer, pattern: %m%n | 583 ns | 583 ns | 1249185 | 
 | Async, int value using MessageBuffer, pattern: %m%n/threads:16 | 978 ns | 7950 ns | 88320 |

swebb2066:
> I don't see a corresponding delete anyplace, am I missing it?

The Node delete is in dispatch at line 574. Raw pointers are used because `std::atomic<T>` only works when T is a small trivially copied type.

swebb2066:
Benchmarks show the simpler `lock_guard<mutex>` synchronization is as performant as an `AtomicList `using `std::atomic<T>.`

swebb2066:
Benchmark measurements were incorrent - did not measure AsyncAppender

swebb2066:
Fixed benchmarks show the simpler lock_guard<mutex> synchronization is more performant then an AtomicList using std::atomic<T>.

PR #340, opened 2024-01-31
Use GCC recommended symbol visibility in the next ABI version

swebb2066:
This PR will reduce the number of exported log4cxx symbols to 5509 from 6737 and reduces other exported symbols to 67 from 108 (in liblog4cxx.so).


swebb2066:
This PR improves the stability of benchmark results
from the following result distribution:
log4cxx-15-benchmark-distribution
to the following result distribution:
log4cxx-16-benchmark-distribution

Issue #344, opened 2024-02-04
FMT macros fail to build with no arguments


rm5248:
I've come back to some code that I'm pretty sure worked before:

```
LOG4CXX_INFO_FMT(logger, "hello");
```

However, GCC 12.2 and clang 14 both fail to compile this code.  I'm fairly certain it compiled properly with GCC 10.2.

The problem is that in our macros we use `__VA_ARGS__` which because there are no arguments seems to cause an error in these compilers.  The workaround for now is to simply use `LOG4CXX_INFO` instead of the `_FMT` series of macros or to simply add an (unused) argument, but this seems a little silly.

It seems that with C++20, we could do `__VA_OPT__(,) __VA_ARGS__`.  Ideally we should be able to support versions before C++20 in a portable way.

Some more info:
https://stackoverflow.com/questions/5891221/variadic-macros-with-zero-arguments
https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

swebb2066:
Support for `LOG4CXX_INFO_FMT(logger, "hello");` was lost a side effect of improving the documentation (see 21f0f12a9bd378cb1b62c0b6218190dcf52ae938).

rm5248:
ah.  That is unfortunate.

I think we can do two things to fix this then:

1. add a preprocessor configuration macro(`LOG4CXX_ALLOW_EMPTY_FMT` or something like that) that enables the old behavior(not used if using C++20 or newer)
2. If C++20 is detected, simply use the `__VA_OPT__(,) __VA_ARGS__` solution in the logging macro.

rm5248:
Looking at it a bit more, the better solution may actually be:

1. For GCC/clang before C++20, use `##__VA_ARGS__`
2. For MSVC before C++20, we may be able to use `__VA_ARGS__` directly("[The traditional Microsoft C++ implementation suppresses a trailing comma if no arguments are passed to the ellipsis.](https://learn.microsoft.com/en-us/cpp/preprocessor/variadic-macros?view=msvc-170)")
3. If a different compiler, use the current behavior until somebody can provide a patch for their specific compiler.
4. If C++ 20, use the `VA_OPT` macro to align with the standard

rm5248:
For my purposes the fix I just merged is fine, since I can upgrade to C++20(I was using C++17).

Is there anything we would like to do in order to make this possible to do in a pre-C++20 environment?

PR #345, opened 2024-02-04
When using C++20, optionally add in comma for varargs

rm5248:
See: #344 

The FMT macros now work with only one argument when using C++20 or greater.

### DIFF
--- a/src/main/include/log4cxx/logger.h
+++ b/src/main/include/log4cxx/logger.h
@@ -2085,6 +2085,12 @@ LOG4CXX_LIST_DEF(LoggerList, LoggerPtr);
 #define LOG4CXX_STACKTRACE
 #endif
 
+#if __cplusplus >= 202002L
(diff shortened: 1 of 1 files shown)


swebb2066:
You can now also undo the [changes in throughput](https://github.com/apache/logging-log4cxx/commit/21f0f12a9bd378cb1b62c0b6218190dcf52ae938#diff-1cf073c63cc93390ba8ee8bc2163a5b77aec896f5187964137de44a57c1c6a47)

rm5248:
Unfortunately since this is a C++20 only fix, that build would fail with earlier versions.

The optional GCC fix as well (`##__VA_ARGS__`) only works if GNU extensions are enabled.  So it works if you compile with `-std=gnu++11` but fails with `-std=c++11`.  There doesn't seem to be a way to test if this is enabled or not at compile time.

Issue #349, opened 2024-02-11
CMake build failure with MultiprocessRollingFileAppender


perhapsmaple:
Hi, I am trying to build log4cxx with the MultiprocessRollingFileAppender using CMake with the following command: 
```
cmake -B build -S . -DLOG4CXX_MULTIPROCESS_ROLLING_FILE_APPENDER=ON -DCMAKE_CXX_STANDARD=17
```

When I build the project, I get the following error:
```
/home/harish/build/original-log4cxx/src/main/cpp/timebasedrollingpolicy.cpp: In member function ‘const string log4cxx::rolling::TimeBasedRollingPolicy::createFile(const string&, const string&, log4cxx::helpers::Pool&)’:
/home/harish/build/original-log4cxx/src/main/cpp/timebasedrollingpolicy.cpp:181:21: error: ‘log4cxx::filesystem’ has not been declared
  181 |         LOG4CXX_NS::filesystem::path path(fileName);
      |                     ^~~~~~~~~~
/home/harish/build/original-log4cxx/src/main/cpp/timebasedrollingpolicy.cpp:182:35: error: ‘path’ was not declared in this scope
  182 |         std::string newFilename = path.filename().string() + szUid + suffix;
      |                                   ^~~~
/home/harish/build/original-log4cxx/src/main/cpp/timebasedrollingpolicy.cpp:182:35: note: suggested alternatives:
In file included from /usr/include/c++/11/filesystem:45,
                 from /home/harish/build/original-log4cxx/build/src/main/include/log4cxx/private/boost-std-configuration.h:9,
                 from /home/harish/build/original-log4cxx/src/main/cpp/timebasedrollingpolicy.cpp:31:
/usr/include/c++/11/bits/fs_path.h:248:9: note:   ‘std::filesystem::__cxx11::path’
  248 |   class path
      |         ^~~~
In file included from /home/harish/build/original-log4cxx/src/main/cpp/timebasedrollingpolicy.cpp:31:
/home/harish/build/original-log4cxx/build/src/main/include/log4cxx/private/boost-std-configuration.h:12:35: note:   ‘{anonymous}::filesystem::path’

   12 |     typedef std::filesystem::path path;
      |                                   ^~~~
/home/harish/build/original-log4cxx/src/main/cpp/timebasedrollingpolicy.cpp:183:21: error: ‘log4cxx::filesystem’ has not been declared
[...]

PR #350, opened 2024-02-11
reorder LOG4CXX_NS cmake variable definition/usage

vopl:


vopl:
https://github.com/apache/logging-log4cxx/issues/349

PR #359, opened 2024-03-07
Benchmark async checks should use AsyncAppender

swebb2066:


rm5248:
I'm not quite sure why that is the case - perhaps there is a lock someplace higher up?  We do lock a mutex when we get all of the appenders in [AppenderAttachableImps](https://github.com/apache/logging-log4cxx/blob/04518326f7a0db532e736a7eaa0f6dd905215fd3/src/main/cpp/appenderattachableimpl.cpp#L80), maybe the multiple threads are blocking on that same mutex?  That could probably be changed to a `shared_mutex`(although I think we removed the check for the `shared_mutex` a little bit ago - might need to add it back in)

swebb2066:
> when we get all of the appenders in [AppenderAttachableImps](https://github.com/apache/logging-log4cxx/blob/04518326f7a0db532e736a7eaa0f6dd905215fd3/src/main/cpp/appenderattachableimpl.cpp#L80), maybe the multiple threads are blocking on that same mutex

It is not that. Removing that mutex (i.e. not copying the vector of appenders, iterating over the member variable directly) does not change the measurements by much.

Only not creating the dispatch thread makes AsyncAppender as fast as other appenders (events are all discarded then).  This suggests the cause is all due to AsyncAppenderPriv::bufferMutex *contention*

rm5248:
Well, everything is effectively serialized down to one thread when appending(because of the mutex), so maybe that actually makes sense?

A lock-free queue might work better, but that's not something I've ever used before.  I think there used to be an option to use the Boot lockfree queue, but I've never used it.  Perhaps we should add that as an option; the public ABI wouldn't change so it would just be a compile-time option.

swebb2066:
> there used to be an option to use the Boost lockfree queue

I have been trying the Boost lockfree queue in the branch `improve_async_appender` but that does not improve performance. The atomic `m_head` pointer in:
```
while (!m_head.compare_exchange_weak(n->next, n, std::memory_order_release))
{
}
```
creates exactly the same content on my x86_64 machine.

PR #361, opened 2024-03-11
Use a compile-time evaluated LocationInfo::calcShortFileName implementation

swebb2066:
Using std::string_view to implement LocationInfo::calcShortFileName results in runtime evaluation on gcc.

Using strrchr to implement LocationInfo::calcShortFileName give a compile time evaluated implementation more often.

To see the generated assembler see [Compiler Explorer](https://godbolt.org/z/nzjjhz4e9)

### DIFF
--- a/src/main/cpp/locationinfo.cpp
+++ b/src/main/cpp/locationinfo.cpp
@@ -45,9 +45,9 @@ LocationInfo::LocationInfo( const char* const fileName1,
 	const char* const methodName1,
 	int lineNumber1 )
 	:  lineNumber( lineNumber1 ),
-	   fileName( fileName1 ),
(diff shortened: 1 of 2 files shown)


rm5248:
I could be misremembering, but I thought that the `string_view` caused MSVC to evaluate the short filename at compile time, and I thought GCC did as well.  MSVC seems to not like to evaluate the short filename at compile-time for some reason.

swebb2066:
> GCC did as well

I discoverd `calcShorthFileName` was not being inlined in my gcc 9.

The MSVC optimiser does not inline if the function calls another - see  [this thread](https://developercommunity.visualstudio.com/t/-declspecdllimport-and-inline-ability-clarificatio/1537756)

PR #363, opened 2024-04-06
Improve  release process repeatability

swebb2066:
This PR aims to make the release process more repeatable by using a scripted job run by CI to generate the release files .


rm5248:
Note: the `APACHE_MAINTAINER` flag does this already: https://github.com/apache/logging-log4cxx/blob/ac7eda0d371a2812a3973b817bba9b6f95c968db/CMakeLists.txt#L265

The one thing that it doesn't do at the moment is sign the artifacts, Volkan has it setup for log4j here I believe: https://github.com/apache/logging-parent/blob/main/.github/workflows/deploy-release-reusable.yaml#L243

swebb2066:
> Note: the APACHE_MAINTAINER flag does this already:

Using CPack would be acceptable if it supported a white-list instead of an ignore-list. At this time, packaging source code seems much simpler when we do not try to use CPack.

> The one thing that it doesn't do at the moment is sign the artifacts

I am not yet knowledgeable enough about signing artifacts to try to do this.

PR #367, opened 2024-04-11
Add 'persist-credentials: false' to all Github checkout actions

swebb2066:
This PR applies a [policy recommended by Apache InfraStructure](https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status#GitHubActionsstatus-Mitigations)

### DIFF
--- a/.github/workflows/abi-compatibility.yml
+++ b/.github/workflows/abi-compatibility.yml
@@ -34,6 +34,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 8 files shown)


swebb2066:
Not sure this is actually necessary. There seems to be a mismatch bewteen the [documentated default](https://github.com/actions/checkout/blob/cd7d8d697e10461458bc61a30d094dc601a8b017/action.yml#L48) and the [code](https://github.com/actions/checkout/blame/cd7d8d697e10461458bc61a30d094dc601a8b017/src/input-helper.ts#L149)

PR #369, opened 2024-04-21
Prefer std::lock_guard when std::unique_lock features are not required

swebb2066:
This PR applies the guidelines in [this highly rated answer on StackOverflow](https://stackoverflow.com/questions/43019598/stdlock-guard-or-stdscoped-lock).

Also, the call site of [std::condition_variable::notify_all](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all) does not need to hold the lock on the same mutex as the one held by the waiting thread(s)

### DIFF
--- a/src/main/cpp/filewatchdog.cpp
+++ b/src/main/cpp/filewatchdog.cpp
@@ -74,11 +74,7 @@ void FileWatchdog::stop()
 {
 	LogLog::debug(LOG4CXX_STR("Stopping file watchdog"));
 	m_priv->interrupted = 0xFFFF;
-
-	{
-		std::unique_lock<std::mutex> lock(m_priv->interrupt_mutex);
-		m_priv->interrupt.notify_all();
-	}
+	m_priv->interrupt.notify_all();
 	m_priv->thread.join();
 }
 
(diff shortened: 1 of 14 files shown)


src/main/cpp/filewatchdog.cpp
```diff
@@ -74,11 +74,7 @@ void FileWatchdog::stop()
 {
 	LogLog::debug(LOG4CXX_STR("Stopping file watchdog"));
 	m_priv->interrupted = 0xFFFF;
-
-	{
-		std::unique_lock<std::mutex> lock(m_priv->interrupt_mutex);
-		m_priv->interrupt.notify_all();
-	}
+	m_priv->interrupt.notify_all();
```
vopl, line 77: There must be a mutex lock between the condition change and cv-notify. Without such a lock the notification can be loosed. Please, take an attention. An atomic condition cannot be a reason for cancellation of a mutex lock, cv logic absolutely requires such a lock.

swebb2066: Thank you for noticing this problem. I have opened #371 to address this.

PR #378, opened 2024-04-29
Improve APRInitiializer interface in next ABI version

swebb2066:
This PR removes the `initialize` method from APRInitiializer. It has not been required since 7fedcdbfd

### DIFF
--- a/src/main/include/log4cxx/helpers/aprinitializer.h
+++ b/src/main/include/log4cxx/helpers/aprinitializer.h
@@ -44,7 +44,11 @@ class FileWatchdog;
 class APRInitializer
 {
 	public:
+#if LOG4CXX_ABI_VERSION <= 15
 		static log4cxx_time_t initialize();
+#else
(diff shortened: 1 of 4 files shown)


src/main/include/log4cxx/helpers/aprinitializer.h
```diff
@@ -44,7 +44,11 @@ class FileWatchdog;
 class APRInitializer
 {
 	public:
+#if LOG4CXX_ABI_VERSION <= 15
 		static log4cxx_time_t initialize();
+#else
+		static log4cxx_time_t getStartTime();
+#endif
```
rm5248, line 51: Instead of having a `#ifdef` here, would it make more sense to just have both methods?

To keep ABI compatibility it's okay to add methods, we just can't remove old ones unless we do an ABI bump.

swebb2066: Yes, I agree in hindsight

PR #379, opened 2024-04-30
Remove duplicated and outdated information from README

swebb2066:
This PR aims to make the Log4cxx README similar to the Log4j2 readme.

Note: The mailing list link was broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 Apache Log4cxx is a logging framework for C++.
 
 For answers to such questions as "What is logging?",
-"What should a logging framework do?' and
+"What should a logging framework do?" and
 "How to do logging properly?"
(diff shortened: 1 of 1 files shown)


rm5248:
This seems a little basic to me.  I think we should probably have the following information at a minimum:

- What log4cxx is(a logging framework for C++).  Could potentially link to the main logging website on what logging is
- License
- Where to file issues(e.g. Github)
- A link to a basic example(could just be the quick start that already exists on the website)

Thoughts?

swebb2066:
Good suggestions. The License is provided by GitHub in the tab.

What do you think about the revised version?

rm5248:
That is at least a reasonable amount of information and points people in the correct direction for most questions.

PR #381, opened 2024-05-06
Join background threads when LOG4CXX_EVENTS_AT_EXIT=on

swebb2066:
This PR ensures Log4cxx threads are joined even when LogManager::shutdown is never called


rm5248:
The updates make me think that maybe we should have a CI action that builds with the events at exit feature configured to be on to make sure it all builds.

swebb2066:
Also need a test case for telnetappender and socketappender

PR #382, opened 2024-05-07
Prevent abnormal termination when telnet appender has no layout

swebb2066:
Sending just the log message seems a reasonable default when a layout is not specified in the configuration. Currently Log4cxx causes abnormal termination in this case.

Also, the number of allowed connections was not documented and could not be configured.

### DIFF
--- a/src/main/include/log4cxx/net/telnetappender.h
+++ b/src/main/include/log4cxx/net/telnetappender.h
@@ -115,6 +115,7 @@ class LOG4CXX_EXPORT TelnetAppender : public AppenderSkeleton
 		Supported options | Supported values | Default value
 		-------------- | ---------------- | ---------------
 		Port | {int} | 23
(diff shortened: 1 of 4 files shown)


src/main/include/log4cxx/net/telnetappender.h
```diff
@@ -80,12 +61,9 @@ class LOG4CXX_EXPORT TelnetAppender : public AppenderSkeleton
 		~TelnetAppender();
 
 		/**
-		This appender requires a layout to format the text to the
-		attached client(s). */
-		bool requiresLayout() const override
-		{
-			return true;
-		}
```
rm5248, line 88: Note: I'm pretty sure changing the inline function to be not inline would cause an ABI version bump, but since this is only used in log4cxx code I think it's safe to do.

PR #392, opened 2024-07-13
Remove APR pool from Logger constructor in next ABI version

swebb2066:
No memory pool should live as long as a Log4cxx logger.


rm5248:
Would it make sense to create a new branch(e.g. `log4cxx-next` or something like that) to do this type of refactoring in?  It might make it a bit cleaner rather than having a bunch of `#if` in the code.

swebb2066:
> Would it make sense to create a new branch


On reflection, I am now inclined to add new methods in 1.3 and generate deprecated warnings on the functions to be removed in the next ABI.

Issue #391, opened 2024-07-12
CustomAppender Example - Class Not Found


sylphrena0:
I'm attempting to create a custom layout to support RFC5424 (see #390), however, I receive a class not found error after compiling my layout and attempting to load the configuration. I attempted to replicate the example extension [provided by documentation](https://logging.apache.org/log4cxx/latest_stable/extending-log4cxx.html). However, after adding `using namespace log4cxx` to resolve the compiler errors with the code, I receive the following error:

```bash
log4cxx: Could not create an Appender. Reported error follows.
log4cxx: Class not found: NullWriterAppender
log4cxx: No appender named [NullAppender] could be found.
Segmentation fault
```

I copied [custom-appender.cpp](https://github.com/apache/logging-log4cxx/blob/4dc3469a8dc2e336346398ee68fc091474e0c996/src/examples/cpp/custom-appender.cpp#L4) verbatim, and copied the `appender` and `appender-ref` to my configuration file. 

Any guidance or suggestions would be appreciated.

I am using `Clang 14.0.6 x86_64-pc-linux-gnu`

rm5248:
Probably the issue is that your naming is incorrect.

The macro `DECLARE_LOG4CXX_OBJECT(NullWriterAppender)` maps to the name of the class, and it is this name(`NullWriterAppender`) that is used to load the class.  

The code and config file go together, so they need to look something like this:
```
class MyAppender : public AppenderSkeleton {
public:
	DECLARE_LOG4CXX_OBJECT(MyAppender)
...
};

In the XML file:
  <appender name="some-name-here" class="MyAppender">
  </appender>
```

Note that the appender does not need to be in the `log4cxx` namespace at all.

sylphrena0:
I am using the configuration from the example, so the naming is correct. For the example, the macro is `IMPLEMENT_LOG4CXX_OBJECT(NullWriterAppender)` and the xml is:

```xml
  <appender
    name="NullAppender"
    class="NullWriterAppender">
    <param
      name="SomeValue"
      value="Nothing" />
  </appender>
  <root>
    <priority value="info" />
    <appender-ref ref="NullAppender" />
  </root>
```

rm5248:
If you've tried that, the only other thing that I can think of is that the class might not be registering properly with log4cxx - this depends slightly on how log4cxx is built/the compiler.

You can try registering the class manually, call the `NullWriterAppender::registerClass()` method directly and then try loading your config file.

[...]

PR #395, opened 2024-07-20
Document a way to avoid SIOF when extending log4cxx

swebb2066:
This PR aims to prevent issues like #391


src/site/markdown/extending.md
```diff
@@ -61,9 +61,14 @@ IMPLEMENT_LOG4CXX_OBJECT(NullWriterAppender)
...
-Log4cxx a way of instansiating our class.  Without these macros, the custom
-class will not work.
+These macros allow Log4cxx to instantiate your class.
+The `DECLARE_LOG4CXX_OBJECT` macro adds method declarations and
+the `IMPLEMENT_LOG4CXX_OBJECT` macro adds implementations.
+The `IMPLEMENT_LOG4CXX_OBJECT` also adds a statically initialized value
+that will register your class name with Log4cxx during program startup.
+To avoid [Static Initialization Order Fiasco](https://en.cppreference.com/w/cpp/language/siof)
+you should place the `IMPLEMENT_LOG4CXX_OBJECT` macro
+in the file that loads the configuration.
```
rm5248, line 71: I don't recall ever having to do this before with GCC on Linux, but perhaps I have not done  a custom appender in a separate cpp file.  Calling the `registerClass()` function seems like a better solution, as it lets you keep all of the implementation in one file.

swebb2066:
The [1.3.0 staging web site](https://logging.staged.apache.org/log4cxx/1.3.0) is available for review.

PR #406, opened 2024-08-12
Update copyright notice

swebb2066:
This PR uses [the text and dates expected by the Apache web crawler](https://whimsy.apache.org/site/project/logging)


rm5248:
Some of those projects are archived, and it mentions Java even though we don't use any Java.  Is this just standard boilerplate from log4j2?

swebb2066:
> Is this just standard boilerplate from log4j2

Yes, I guess it is Apache Logging boilerplate. Does it matter?

Issue #370, opened 2024-04-21
Use Logging Services common pages


vy:
I have revamped the [Download](https://logging.apache.org/download.html), [Support](https://logging.apache.org/support.html), and [Security](https://logging.apache.org/security.html) pages of Logging Services (LS) such that they now cover all LS projects. In the Log4j website, I have wired all associated links to there and I have replaced all overlapping content to point to there. Would you consider implementing the following changes for the left menu of the Log4cxx website?

1. Replace _"Community"_ and all sublinks/pages under it with a _"Support"_ link pointing to the LS Support page
1. Point _"Get > Download"_ to the LS Download page
1. Add a _"Security"_ link pointing to the LS Security page

vy:
> As far as I can see, those three pages do not seem to have anything that is relevant to a Log4cxx user.

Pages I cited are implemented in a project agnostic way. That is,

* [Download](https://logging.apache.org/download.html) points to locations where users can download artifacts **of all projects** and shows how to verify them
* [Support](https://logging.apache.org/support.html) shares official support channels, i.e., mailing lists, GitHub Discussions, GitHub Issues **of all projects**. (Search for `log4cxx` in that page for Log4cxx-specific entries.) It guides users using _any_ LS project on how to get support.
* [Security](https://logging.apache.org/security.html) page (which Log4cxx websites currently doesn't have an equivalent right now!) guides users on how to get support on security issues, and shares VDR files and CVE lists covering **all projects**.

Hence, I think the content covers all Download-Support-Security needs of LS projects, including Log4cxx. If you think something is falling short from a Log4cxx perspective, I would be more than happy to accommodate that.

Note that, yesterday [Log4net](https://logging.apache.org/log4net/) also migrated all their Download-Support-Security pages too.

> I notice the links to these pages are available from the [Apache Logging Services main page](https://logging.apache.org/). Would not that be adequate?

The goal is to

* Remove the maintenance burden of these pages for each project individually
* Create a coherent Download-Support-Security experience for users of all LS projects

Log4j, Log4j Scala, Log4j Kotlin, Log4Net, etc. all have already migrated. Log4cxx is the only project left out.

[...]

PR #411, opened 2024-09-17
Google OSS-Fuzz integration

vy:
As a deliverable of apache/logging-log4j2#2891 and apache/logging-log4j2#2892, this PR implements fuzz tests along with Google OSS-Fuzz integration. See the added `fuzzing.md` for details.

**Note:** Review request has been submitted in [this `dev@` post](https://lists.apache.org/thread/0yqh7ffr5gk031hszlp29oxzr3l57y1d).

### DIFF
--- a/.github/workflows/log4cxx-ubuntu.yml
+++ b/.github/workflows/log4cxx-ubuntu.yml
@@ -34,6 +34,7 @@ jobs:
             odbc: OFF
             multithread: OFF
             exitevents: OFF
+            fuzzers: OFF
           - name: ubuntu20-clang
             os: ubuntu-20.04
             cxx: clang++
@@ -42,6 +43,7 @@ jobs:
             odbc: ON
             multithread: OFF
             exitevents: OFF
+            fuzzers: ON
           - name: ubuntu22-gcc
             os: ubuntu-22.04
             cxx: g++
@@ -50,6 +52,7 @@ jobs:
             odbc: OFF
             multithread: ON
             exitevents: ON
+            fuzzers: OFF
           - name: ubuntu22-clang
             os: ubuntu-22.04
             cxx: clang++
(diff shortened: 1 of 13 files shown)


rm5248:
This is looking good!  I won't have a chance to run this myself for a few days, but it looks like I can run it manually?  What about running it as part of our github actions?

vy:
> This is looking good! I won't have a chance to run this myself for a few days, but it looks like I can run it manually?

No worries. There are still some rough edges at the OSS-Fuzz side, hence I will be polishing it a bit more. Note that the PR is marked as `Draft`. [Earlier @swebb2066 had explicitly told me to not submit features via GitHub without a mailing list discussion](https://github.com/apache/logging-log4cxx/issues/370#issuecomment-2074783658). I will do that when the PR is ready to be reviewed.

> What about running it as part of our github actions?

That is an excellent question! :star_struck: Fuzzers are intended to be run for (preferably) long periods of time and regularly. To speed things up, you need to save your state (called **corpus**), and restore it in the next run. When a fuzzers fails, you need to take a snapshot of the context (preferably, somewhere not disclosed to public for security reasons, because you might have just stumbled upon a vulnerability) to allow reproduction, and continuously verify these reproductions as code base changes. This sophisticated pipeline also needs an administrative UI and reporting features. We can implement such a GitHub Actions workflow, leverage either ASF Subversion repository or some private GitHub repository for storage, etc. But this is a pretty big task! Good news is, this is what OSS-Fuzz exactly does! :sweat_smile: A beefy cluster continuously fuzzing, storing its state & findings to a GCP bucket, and giving visibility to its pipeline state using a web page.

Last 2 months I've learned a lot while trying to integrate Log4j and Log4cxx to OSS-Fuzz. If I would do it all over again, and have more time+financial budget, I'd go the GHA route. This would not only put us in complete control, but also result in a product that any GitHub project can use. But alas there is a deadline I need to deliver this task.

rm5248 (approved):
This looks good to me.  Once I installed docker and followed the directions, I was able to create the fuzzed data and run it.  I let it run for a few hours before killing it; I'll let it run all day tomorrow and see what the results are.

Kindly squash merge when merging in.

PR #415, opened 2024-10-15
Capture stack trace when ucrt-x86_64 asyncappendertestcase blocks

swebb2066:


swebb2066:
Now that the behaviour of asynchappender multiThreadTest is understood and fixed in #402, this workflow is no longer needed

PR #446, opened 2025-01-11
The EXPAT library dependency is only required in a statically linked Log4cxx

swebb2066:
This PR aims to clarify (for package managers) when a direct dependency on EXPAT library and the Threads library is required.


CMakeLists.txt
```diff
@@ -327,7 +322,9 @@ message(STATUS "  Using syslog.h .................. : ${HAS_SYSLOG}")
 endif()
 message(STATUS "  TelnetAppender .................. : ${LOG4CXX_NETWORKING_SUPPORT}")
 message(STATUS "  NTEventLogAppender .............. : ${HAS_NT_EVENTLOG_APPENDER}")
+if(WIN32)
```
rm5248, line 325: should this be moved up one line?  The NTEventLogAppender would only be available on Windows.

swebb2066: On the basis that showing an unavailable option is not helpful, yes NTEventLogAppender status should only be shown on WIN32

PR #456, opened 2025-01-18
Add PatternParser fuzzer

AdamKorcz:
Adds a fuzzer for `PatternParser::parse()`.

### DIFF
--- a/src/fuzzers/cpp/CMakeLists.txt
+++ b/src/fuzzers/cpp/CMakeLists.txt
@@ -15,7 +15,12 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


vy:
Wouldn't it be better to only update the pattern in `PatternLayoutFuzzer.properties` (which is employed by `PatternLayoutFuzzer.cpp`) instead of creating a new fuzzer?

swebb2066:
Advice from [libFuzzer documentation](https://llvm.org/docs/LibFuzzer.html#id23) is:

> Usually, the narrower the target the better. E.g. if your target can parse several data formats, split it into several targets, one per format

PR #460, opened 2025-01-23
Add DOM Configurator fuzzer

AdamKorcz:
Adds a fuzzer that targets `DOMConfigurator::configure()`.

### DIFF
--- a/src/fuzzers/cpp/DOMConfiguratorFuzzer.cpp
+++ b/src/fuzzers/cpp/DOMConfiguratorFuzzer.cpp
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "stdint.h"
+#include <fuzzer/FuzzedDataProvider.h>
+#include <log4cxx/xml/domconfigurator.h>
+
+using namespace log4cxx;
+using namespace log4cxx::helpers;
+using namespace log4cxx::xml;
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	char filename[9] = "conf.xml";
+
+	FILE *fp = fopen(filename, "wb");
+	if (!fp) {
+		return 0;
+	}
+	fwrite(data, size, 1, fp);
+	fclose(fp);
+	DOMConfigurator::configure(filename);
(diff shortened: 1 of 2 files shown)


src/fuzzers/cpp/DOMConfiguratorFuzzer.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+using namespace log4cxx;
+using namespace log4cxx::helpers;
+using namespace log4cxx::xml;
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+	char filename[9] = "conf.xml";
+
+	FILE *fp = fopen(filename, "wb");
+	if (!fp) {
```
swebb2066, line 30: Did you mean to provide/generate `conf.xml`? Otherwise this is always going to fail.

AdamKorcz: `fopen()` creates the file if it doesn't exist, if that is what you are concerned of?

swebb2066: Testing DOMConfigurator::configure() with an empty file should be just a unit test.

A fuzz test of xml content would be a better use of oss-fuzz resources.

rm5248: It's writing the file a few lines below, so I believe this is properly writing out the data to a file and then having the `DOMConfigurator` parse the file that was just written.

swebb2066: What do you expect this fuzz test to achieve?

Have you considered mimicing the fuzz test in [libexpat](https://github.com/libexpat/libexpat/tree/master/expat/fuzz)?

swebb2066: For OSS-fuzz to deeply investigate the DOMConfigurator code, it would need to generate mostly valid log4cxx configuration which would be extremely unlikely, as admitted by this concluding comment in [this blog post](https://lcamtuf.blogspot.com/2014/11/pulling-jpegs-out-of-thin-air.html)

> In practical terms, this means that afl-fuzz won't have as much luck "inventing" PNG files or non-trivial HTML documents from scratch

AdamKorcz: > For OSS-fuzz to deeply investigate the DOMConfigurator code, it would need to generate mostly valid log4cxx configuration which would be extremely unlikely, as admitted by this concluding comment in [this blog post](https://lcamtuf.blogspot.com/2014/11/pulling-jpegs-out-of-thin-air.html)
> 
> > In practical terms, this means that afl-fuzz won't have as much luck "inventing" PNG files or non-trivial HTML documents from scratch

libFuzzer will be able to explore the code because of its coverage-guided features. We can help it too by adding a seed and a dictionary which will greatly improve the fuzzers reach from the get go. Once the fuzzer is merged, I can add that to the OSS-Fuzz build files.

> What do you expect this fuzz test to achieve?

The fuzzer tests [mostly XML parsing.](https://github.com/apache/logging-log4cxx/blob/master/src/main/cpp/domconfigurator.cpp). The DOM configurator has 1000+ LoC which would be great to have fuzz coverage of. I don't see a reason to not test it.

> Have you considered mimicing the fuzz test in [libexpat](https://github.com/libexpat/libexpat/tree/master/expat/fuzz)?

No. Could you elaborate on the advantages of that?

> Testing DOMConfigurator::configure() with an empty file should be just a unit test.

The file will not be empty. The fuzzer writes the test case to the file and then invokes the DOM Configurator so that it parses it.

PR #461, opened 2025-01-24
Add Time based rolling policy fuzzer

AdamKorcz:
Adds a fuzzer that tests `TimeBasedRollingPolicy->rollover()`.

### DIFF
--- a/src/fuzzers/cpp/TimeBasedRollingPolicyFuzzer.cpp
+++ b/src/fuzzers/cpp/TimeBasedRollingPolicyFuzzer.cpp
@@ -52,8 +52,6 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     tbrp->activateOptions(pool);
     rfa->setRollingPolicy(tbrp);
     rfa->activateOptions(pool);
-    rfa->setBufferSize(fdp.ConsumeIntegral<int>());
-    rfa->activateOptions(pool);
     rfa->setBufferedSeconds(fdp.ConsumeIntegral<int>());
     rfa->activateOptions(pool);
     LoggerPtr logger = LogManager::getLogger("org.apache.log4j.TimeBasedRollingTest");
--- a/src/fuzzers/cpp/TimeBasedRollingPolicyFuzzer.cpp
+++ b/src/fuzzers/cpp/TimeBasedRollingPolicyFuzzer.cpp
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


src/fuzzers/cpp/TimeBasedRollingPolicyFuzzer.cpp
```diff
@@ -0,0 +1,80 @@
...
+           tbrp->setFileNamePattern(LogString(LOG4CXX_STR("fuzz-%d{" DATE_PATTERN "}.gz")));
+    } else {
+           tbrp->setFileNamePattern(LogString(LOG4CXX_STR("fuzz-%d{" DATE_PATTERN "}.zip")));
+    }
+    rfa->setFile(LOG4CXX_STR(LOG4CXX_STR("test.log")));
+    
+    tbrp->activateOptions(pool);
+    rfa->setRollingPolicy(tbrp);
+    rfa->activateOptions(pool);
+    rfa->setBufferSize(fdp.ConsumeIntegral<int>());
```
swebb2066, line 55: This will violate an undocumented but assumed precondition that a buffer of this size can be allocated by the system.

swebb2066 (commented):
Preconditions should be specified in the interface but are not yet being part of the C++ language.

Do you agree that the caller has a responseability to honour preconditions?

src/fuzzers/cpp/TimeBasedRollingPolicyFuzzer.cpp
AdamKorcz: Thanks. What would a generous but still reasonable max size be here? Could we do 10,000?

AdamKorcz:
> Do you agree that the caller has a responseability to honour preconditions?

Yes.

src/fuzzers/cpp/TimeBasedRollingPolicyFuzzer.cpp
swebb2066: I suggest leaving it as a default. A reasonable precondition would say "big enough for a few logging events and not bigger than the system handle"

rm5248: Looking at the code, it seems that it is not actually allocating the amount of memory as defined by `setBufferSize`.  It will expand up to the limit: https://github.com/apache/logging-log4cxx/blob/da0313eb333e4f76ed7ade48075025f64f9c29e9/src/main/cpp/bufferedwriter.cpp#L67

So a random number here probably doesn't hurt, although I don't think it would actually cause any bugs either.

swebb2066: In a 32-bit architecture, when buffersize larger than the system can handle, `m_priv->buf.append(str)` will report out-of-memory. In a 64-bit architecture, the system will slow to a crawl paging to disk before it gets to report out-of-memory.

AdamKorcz: Should be fixed with https://github.com/apache/logging-log4cxx/pull/461/commits/3febd7a0c0a43085e7b5cfdf27adcfb856fc8d8b.

PR #465, opened 2025-01-29
build fuzzers with utf-8 and wchar_t

AdamKorcz:
Set up the CMake file so it builds each fuzzer with utf-8 and wchar_t.

I have had to remove the `LOG4CXX_STR` and use the Transcoder instead to create `LogString` variables.

### DIFF
--- a/src/fuzzers/bash/oss-fuzz-build.sh
+++ b/src/fuzzers/bash/oss-fuzz-build.sh
@@ -43,9 +43,8 @@ declare -a encodings=("utf-8" "wchar_t")
 
 for encoding in "${encodings[@]}"
 do
-  echo heeeeere
   # Build the project
-  mkdir -p build && cd $_
+  mkdir -p build-"${encoding}" && cd $_
   cmake \
     -DBUILD_SHARED_LIBS=OFF \
     -DBUILD_TESTING=OFF \
@@ -58,5 +57,5 @@ do
   # Copy executables & resources
   find src/fuzzers/cpp -maxdepth 1 -executable -type f -exec cp -v {} "$outputDir/" \;
   cp ../src/fuzzers/resources/* "$outputDir/"
-  cd .. && rm -r build
+  cd .. && rm -r build-"${encoding}"
 done
--- a/src/fuzzers/bash/oss-fuzz-build.sh
+++ b/src/fuzzers/bash/oss-fuzz-build.sh
@@ -38,16 +38,25 @@ mkdir -p "$outputDir"
 # Switch to the project directory (by referencing from the script directory)
 cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && cd ../../..
 
-# Build the project
-mkdir -p build && cd $_
-cmake \
-  -DBUILD_SHARED_LIBS=OFF \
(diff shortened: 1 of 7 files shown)


src/fuzzers/cpp/CMakeLists.txt
```diff
@@ -49,31 +49,34 @@ else()
...
-      ${APR_LIBRARIES}
-      ${APR_SYSTEM_LIBS})
-  if(WIN32)
-      set_target_properties(${PROGRAM_NAME}
-        PROPERTIES
-          VS_DEBUGGER_ENVIRONMENT "PATH=${ESCAPED_PATH}"
-          VS_DEBUGGER_WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
-          FOLDER Fuzzers)
-  endif()
+  foreach(logchar IN LISTS FUZZ_CHARS)
```
swebb2066, line 52: LOG4CXX_CHAR must be set at the build directory level. Its value is [written into log4cxx.h](https://github.com/apache/logging-log4cxx/blob/b0775e7d4208d39007c33f7bc04f61eacff7979e/src/main/include/log4cxx/log4cxx.h.in#L48)

AdamKorcz: Thanks. Have updated the PR.

src/fuzzers/bash/oss-fuzz-build.sh
```diff
@@ -38,16 +38,25 @@ mkdir -p "$outputDir"
...
-
-# Copy executables & resources
-find src/fuzzers/cpp -maxdepth 1 -executable -type f -exec cp -v {} "$outputDir/" \;
-cp -v ../src/fuzzers/resources/* "$outputDir/"
+# Build a fuzzer with each encoding
+declare -a encodings=("utf-8" "wchar_t")
+
+for encoding in "${encodings[@]}"
+do
+  echo heeeeere
```
rm5248, line 46: debug message?

src/fuzzers/bash/oss-fuzz-build.sh
```diff
@@ -38,16 +38,25 @@ mkdir -p "$outputDir"
...
+  # Build the project
+  mkdir -p build && cd $_
```
rm5248, line 48: I'm not sure if this will cause a problem in this case, but sometimes cmake needs to clear the build dir between runs when changing certain variables(I think if the variable is cached).  Do we need separate build directories and/or to clear out the build directory between runs?

AdamKorcz: Yes, I believe you are right. I have renamed the cmake build dirs in https://github.com/apache/logging-log4cxx/pull/465/commits/c8cb362b03f1e3bf6b9b0af071cd8c3ba43efd79

src/fuzzers/bash/oss-fuzz-build.sh
AdamKorcz: Yes, thanks for catching! Fixed in https://github.com/apache/logging-log4cxx/pull/465/commits/c8cb362b03f1e3bf6b9b0af071cd8c3ba43efd79

PR #470, opened 2025-02-01
Fix possible invalid cast

rm5248:
If a user attempts to load a config file with an improper level set such that a non-level class is attempted to be used, throw an exception instead of accessing invalid memory.

### DIFF
--- a/src/main/cpp/optionconverter.cpp
+++ b/src/main/cpp/optionconverter.cpp
@@ -343,8 +343,13 @@ LevelPtr OptionConverter::toLevel(const LogString& value,
 
 	try
 	{
-		Level::LevelClass& levelClass =
-			(Level::LevelClass&)Loader::loadClass(clazz);
+		// Note: the dynamic_cast could fail across DLL boundaries.
+		// However, without the dynamic_cast a poorly formed XML file
+		// could attempt to load an invalid class as a filter, causing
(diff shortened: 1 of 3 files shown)


src/main/cpp/optionconverter.cpp
```diff
@@ -343,8 +343,13 @@ LevelPtr OptionConverter::toLevel(const LogString& value,
...
+		// Note: the dynamic_cast could fail across DLL boundaries.
+		// However, without the dynamic_cast a poorly formed XML file
+		// could attempt to load an invalid class as a filter, causing
+		// a crash.  If it can't be converted, a std::bad_cast should be
+		// thrown(and caught by the exception handler below)
+		const Level::LevelClass& levelClass =
+			dynamic_cast<const Level::LevelClass&>(Loader::loadClass(clazz));
 		return levelClass.toLevel(levelName);
 	}
 	catch (ClassNotFoundException&)
```
swebb2066, line 355: The final exception handler could be more specific to catch std::bad_cast, and I think it is always useful to include information in logged warnings.
```
    catch (const std::exception& ex)
	{
		LogLog::warn(
			LOG4CXX_STR("class [") + clazz + LOG4CXX_STR("], level [") + levelName +
			LOG4CXX_STR("] conversion) failed."), ex);
	}

```

PR #475, opened 2025-02-03
Restrict pattern layout min/max field width specifiers to 3 digits

swebb2066:
This PR proposes adding documentation and checks to avoid integer-overflow bugs.


swebb2066:
Fixes [OSS-Fuzz 393666942](https://issues.oss-fuzz.com/issues/393666942) and [OSS-Fuzz 393411535](https://issues.oss-fuzz.com/issues/393411535)

PR #477, opened 2025-02-05
Prevent a fault when an option name has an emedded NUL character

swebb2066:
This PR addresses [this OSS-Fuzz issue](https://issues.oss-fuzz.com/issues/394331188)

### DIFF
--- a/src/main/cpp/stringhelper.cpp
+++ b/src/main/cpp/stringhelper.cpp
@@ -41,7 +41,7 @@ bool StringHelper::equalsIgnoreCase(const LogString& s1, const logchar* upper, c
 		++lower;
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/cpp/stringhelper.cpp
```diff
@@ -30,17 +30,18 @@ using namespace LOG4CXX_NS::helpers;
 
 bool StringHelper::equalsIgnoreCase(const LogString& s1, const logchar* upper, const logchar* lower)
 {
-	for (LogString::const_iterator iter = s1.begin();
-		iter != s1.end();
-		iter++, upper++, lower++)
+	for (auto& item : s1)
```
rm5248, line 33: perhaps make this `const auto&` to match the `const_iterator` from before

PR #488, opened 2025-03-16
Allow for use of either Qt5 or Qt6

rm5248:
Since Qt `#defines emit`, the `LogLog` class fails to compile properly as the function prototype no longer exists.

Oddly this works fine in Qt5, but it breaks with Qt6.  We can work around it with defining `QT_NO_KEYWORDS`, but this would require anybody who links with us and with Qt to define this as well, which is not a good requirement for a library.

### DIFF
--- a/.github/workflows/log4cxx-ubuntu.yml
+++ b/.github/workflows/log4cxx-ubuntu.yml
@@ -31,6 +31,7 @@ jobs:
             cxx: g++
             fmt: OFF
             qt: ON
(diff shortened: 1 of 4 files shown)


rm5248:
Note: if removing the `emit` function prototype breaks the ABI(I don't think it should since it's `private static` and is only inside of the `LogLog` class), we can always wrap it in a preprocessor macro that is only defined when building the library itself.

PR #490, opened 2025-03-21
Update links to official support channels

swebb2066:
This PR updates the links on the support page to reflect that the only supported mail list is dev@logging.apache.org

### DIFF
--- a/src/site/markdown/community.md
+++ b/src/site/markdown/community.md
@@ -33,10 +33,9 @@ some issues simply aren't bugs in the code base of the project or it's build too
 but may have a lot of different other reasons.
 
 If things really only don't work and you don't have any clue why,
-please use the [official support channels],
-i.e., [usage mailing list](mailto:log4cxx-user@logging.apache.org),
-[developer mailing lists](mailto:log4cxx-dev@logging.apache.org)
-or [GitHub Discussions] instead.
+please use the [official support channels] instead,
+i.e., [GitHub Discussions]
+or [developer mailing lists].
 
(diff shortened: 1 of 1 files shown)


rm5248:
should we add notes on the security threat model?  It's being updated on the main logging site, I don't know if we want to reiterate it: https://github.com/apache/logging-site/pull/10

ppkarwasz:
> should we add notes on the security threat model? It's being updated on the main logging site, I don't know if we want to reiterate it: [apache/logging-site#10](https://github.com/apache/logging-site/pull/10)

It would be nice if you can look at that PR. I must admit that Log4cxx is the project I am the least familiar with. For example I don't know if the configuration can be changed at runtime and if there is a chance of log event loss during reconfiguration.

swebb2066:
> I don't know if the configuration can be changed at runtime and if there is a chance of log event loss during reconfiguration.

Yes, configuration can be changed at runtime. I do not know if log event loss during reconfiguration is prevented.

rm5248:
> > I don't know if the configuration can be changed at runtime and if there is a chance of log event loss during reconfiguration.
> 
> Yes, configuration can be changed at runtime. I do not know if log event loss during reconfiguration is prevented.

It probably depends on if the application is single or multi-threaded, and the configuration that you have.  If it's single threaded and you're only logging to a file, then a reconfiguration wouldn't lose any log messages.

I think the core is properly mutexed, so when an appender gets reconfigured it will block any new events from coming in until the configuration is applied.

PR #491, opened 2025-03-26
Ensure log events are sent to an appender during a reconfiguration event

swebb2066:
This PR addresses a race condition where an active thread sends an event to a Logger after the reconfiguration thread has removed all appenders but before the new appender list is established.

The new *replaceAppender* methods block event producing threads while the list of appenders is updated.

The modified test case [errorhandler test](https://github.com/apache/logging-log4cxx/compare/reconfiguration_synchronization?expand=1#diff-8d655beadcefc87795cfe34d8c7796085a484cb0b3ff8e5cc4d51d4145951ba3) previously used an invalid configuration to trigger the fallback handler. This fallback process (Installing the FALLBACK appender in place of the faulty PRIMARY appender) was happening during configuration, but the PRIMARY appender (with a NULL writer) was later being added to the root logger. After configuration the root logger had two appenders, one of which had a a NULL writer.

I believe the intent of the errorhandler test is in checking the fallback handling on any error, not specifically a faulty configuration.

The behaviour of Log4cxx when the configuration file is invalid should ideally be to leave the current configuration unchanged, as happens when the xml is not parsable. However doing this for other types of configuration errors is a larger job than I am attempting here.

### DIFF
--- a/src/main/cpp/appenderattachableimpl.cpp
+++ b/src/main/cpp/appenderattachableimpl.cpp
@@ -167,4 +167,37 @@ void AppenderAttachableImpl::removeAppender(const LogString& name)
 	}
 }
 
+bool AppenderAttachableImpl::replaceAppender(const AppenderPtr& oldAppender, const AppenderPtr& newAppender)
+{
+	bool found = false;
+	if (m_priv && oldAppender && newAppender)
+	{
+		auto oldName = oldAppender->getName();
+		std::lock_guard<std::mutex> lock( m_priv->m_mutex );
+		auto it = std::find_if(m_priv->appenderList.begin(), m_priv->appenderList.end()
+			, [&oldName](const AppenderPtr& appender) -> bool
+			{
+				return oldName == appender->getName();
+			});
+		if (it != m_priv->appenderList.end())
+		{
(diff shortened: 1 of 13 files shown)


rm5248:
> I believe the intent of the errorhandler test is in checking the fallback handling on any error, not specifically a faulty configuration.


There are at least two separate problems that I am aware of that the fallback error handler is used for: if the appender becomes invalid when reconfiguring, and if it becomes invalid during runtime.

The second possibility is discussed somewhat in https://github.com/apache/logging-log4cxx/pull/62, where the zip/gzip compress actions can fail under certain circumstances outside of our control.  If appending to the file and writing to it fails, the error handler may be called in order to swap out the appender.

swebb2066:
NB: There is a change in bahavior in FallbackErrorHandler.cpp. The *replaceAppender* call does not add the fallback appender to a logger without a refence to the failed appender.

I believe this to be more in line with the [documented behaviour on FallbackErrorHandler::setLogger](https://logging.apache.org/log4cxx/1.4.0/classlog4cxx_1_1varia_1_1FallbackErrorHandler.html#a69d8bbc93046e806a40167f1dc414499) and less surprising than the current behaviour.

PR #492, opened 2025-03-29
Update CI Ubuntu versions

swebb2066:
This PR moves CI actions off Ubuntu 20 which [soon will not be supported](https://github.com/actions/runner-images/issues/11101)

### DIFF
--- a/.github/workflows/abi-compatibility.yml
+++ b/.github/workflows/abi-compatibility.yml
@@ -40,9 +40,7 @@ jobs:
     - name: 'Configure Dependencies - Ubuntu'
       run: |
         sudo apt-get update
-        sudo apt-get install -y gcc-9 libapr1-dev libaprutil1-dev elfutils vtable-dumper universal-ctags
-        # note: sqlext.h exists on github VM, purge for now as we don't link correctly...
-        sudo apt-get purge unixodbc-dev
+        sudo apt-get install -y g++-9 libapr1-dev libaprutil1-dev elfutils vtable-dumper universal-ctags
 
     - name: 'run CMake'
       run: |
@@ -53,13 +51,11 @@ jobs:
         cmake --build .
 
     - name: 'copy compat report'
-      if: always()
       run: |
         cp $(find . -name compat_report.html) .
         cp $(find . -name new-abi.dump) .
 
     - uses: actions/upload-artifact@v4
-      if: always()
       with:
         name: abi_info
         path: |
--- a/.github/workflows/abi-compatibility.yml
+++ b/.github/workflows/abi-compatibility.yml
@@ -51,11 +51,13 @@ jobs:
         cmake --build .
 
     - name: 'copy compat report'
+      if: always()
       run: |
         cp $(find . -name compat_report.html) .
         cp $(find . -name new-abi.dump) .
 
     - uses: actions/upload-artifact@v4
+      if: always()
       with:
         name: abi_info
         path: |
--- a/.github/workflows/abi-compatibility.yml
+++ b/.github/workflows/abi-compatibility.yml
@@ -26,7 +26,7 @@ on: [push, pull_request]
 jobs:
   job:
     name: abi-check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     timeout-minutes: 38
     strategy:
       fail-fast: false
@@ -40,7 +40,7 @@ jobs:
     - name: 'Configure Dependencies - Ubuntu'
       run: |
         sudo apt-get update
-        sudo apt-get install -y libapr1-dev libaprutil1-dev elfutils vtable-dumper universal-ctags
+        sudo apt-get install -y gcc-9 libapr1-dev libaprutil1-dev elfutils vtable-dumper universal-ctags
         # note: sqlext.h exists on github VM, purge for now as we don't link correctly...
         sudo apt-get purge unixodbc-dev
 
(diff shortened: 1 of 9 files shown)


swebb2066:
The src/main/abi-symbols/abi.dump produced from g++ 9.4 contains
```
                          '14545351' => {
                                          'Header' => 'propertysetter.h',
                                          'Line' => '62',
                                          'Memb' => {
                                                      '0' => {
                                                               'access' => 'protected',
                                                               'name' => 'obj',
                                                               'offset' => '0',
                                                               'type' => '648233'
                                                             }
                                                    },
                                          'Name' => 'log4cxx::config::PropertySetter',
                                          'NameSpace' => 'log4cxx::config',
                                          'Size' => '16',
                                          'Type' => 'Class'
                                       },
and
                          '648233' => {
                                        'BaseType' => '565797',
                                        'Header' => 'object.h',
                                        'Line' => '114',
                                        'Name' => 'log4cxx::helpers::ObjectPtr',
                                        'NameSpace' => 'log4cxx::helpers',
                                        'Size' => '16',
                                        'Type' => 'Typedef'
                                      },

```
The new-abi.dump produced from g++ 9.5 contains
```
                          '4837555' => {
                                         'Header' => 'propertysetter.h',
                                         'Line' => '62',
                                         'Memb' => {
                                                     '0' => {
                                                              'access' => 'protected',
                                                              'name' => 'obj',
                                                              'offset' => '0',
                                                              'type' => '455809'
                                                            }
                                                   },
                                         'Name' => 'log4cxx::config::PropertySetter',
                                         'NameSpace' => 'log4cxx::config',
                                         'Size' => '16',
                                         'Type' => 'Class'
                                       },
and
                          '455809' => {
                                        'BaseType' => '390060',
                                        'Header' => 'object.h',
                                        'Line' => '114',
                                        'Name' => 'log4cxx::helpers::ObjectPtr',
                                        'NameSpace' => 'log4cxx::helpers',
                                        'Type' => 'Typedef'
                                      },
```

rm5248:
Are we getting a false positive with the ABI check?  I can take a look at it later.

swebb2066:
Compiling this branch with g++ 11.4.0 produces a new-abi.dump containing (for example):
```
                            '18461936' => {
                                            'Class' => '11845076',
                                            'Header' => 'hierarchy.h',
                                            'MnglName' => '_ZN7log4cxx9Hierarchy18ensureIsConfiguredESt8functionIFvvEE',
                                            'Param' => {
                                                         '0' => {
                                                                  'name' => 'this',
                                                                  'type' => '18474272'
                                                                },
                                                         '1' => {
                                                                  'name' => 'configurator',
                                                                  'type' => '2274699'
                                                                }
                                                       },
                                            'Reg' => {
                                                       '1' => 'rsi'
                                                     },
                                            'Return' => '1',
                                            'ShortName' => 'ensureIsConfigured',
                                            'Source' => 'hierarchy.cpp',
                                            'SourceLine' => '302',
                                            'Virt' => 1,
                                            'VirtPos' => '6'
                                          },
```
Compiling the rel/v1.2.0 branch with g++ 11.4.0 produces a new-abi.dump containing:
```
                            '18603406' => {
                                            'Class' => '11930657',
                                            'Header' => 'hierarchy.h',
                                            'MnglName' => '_ZN7log4cxx9Hierarchy18ensureIsConfiguredESt8functionIFvvEE',
                                            'Param' => {
                                                         '0' => {
                                                                  'name' => 'this',
                                                                  'type' => '18615999'
                                                                },
                                                         '1' => {
                                                                  'name' => 'configurator',
                                                                  'type' => '2331741'
                                                                }
                                                       },
                                            'Reg' => {
                                                       '0' => 'rbx',
                                                       '1' => 'rbp'
                                                     },
                                            'Return' => '1',
                                            'ShortName' => 'ensureIsConfigured',
                                            'Source' => 'hierarchy.cpp',
                                            'SourceLine' => '310',
                                            'Virt' => 1,
                                            'VirtPos' => '6'
                                          },
```

rm5248:
What compiler did you generate the new ABI dump file with?

Taking a look at the compat report, the problems that it's showing are real problems with the ABI, but I'm pretty sure they're false positives, although it does depend on what compiler is being used.

Testing with g++ 10.2, I get only 5 problems that all appear to be false positives:
* It thinks RootLogger() constructor changed, which the signature did not
* it thinks Transcoder::decode() changed, but it did not
* It thinks that the `obj` field in PropertySetter changed from 16 bytes to 0 bytes

It seems that ubuntu 20 is using g++9.3, which uses C++14 by default.  g++ versions 11 and above use C++17 by default.

Some of the errors(e.g. `The parameter pool1 became passed in r12 register instead of rbp.`) might be due to the different C++ standard being used, although if the name mangling is the same I would expect the calling convention to be the same.  It is possible to version symbols with gcc though, so there could be an invisible "works with C++ version X" version being added to the symbol.

I'm pretty sure the ABI check for C++ is only useful if the same C++ standard is used between all of the checks, which is why the abi-check build is only on a specific version of ubuntu(and not latest).  From the ABI compliance checker page:
> The ABI dump can be used to create a snapshot of a library ABI in the particular environment and then compare it with any other state of the ABI changed due to changes in the environment (compiler version, external libraries, etc.) or changes in the library API (header files). 

This would imply to me that the change in compiler version and libstdc++ version are enough to make the ABI check fail; we're concerned with the last part(changes in the library API).

swebb2066:
> What compiler did you generate the new ABI dump file with?

I used g++ 11.4.0 to generate the new ABI dump file using the rel/v1.2.0 source code, i.e. the same version used to build this branch. This was to ensure a change in compiler version was not the issue.

Using gdb to disassemble a call to `log4cxx::Hierarcchy::ensureIsConfigured` in rel/v1.2.0 and this branch shows the same registers (rsi, rdi) are used. 

So the problem is abi-dumper.pl not correctly processing the output of `eu-readelf` for recent g++ versions.

ppkarwasz:
Is there a way to make the `abi.dump` in this repository more easily comparable using `diff`? Sorting the entries by `MnglName` should help, right?

swebb2066:
> Is there a way to make the abi.dump in this repository more easily comparable using diff? Sorting the entries by MnglName should help, right?

No, it is full of magic numbers, which change with every build.

The [abi-compliance-checker](https://github.com/lvc/abi-compliance-checker) perl project produces a html summary of the differences which is available from a Github link under _Run actions/upload-artifact_ step in the abi-check Github action log.

PR #495, opened 2025-04-16
Add SO_REUSEADDR to APRServerSocket

michieldwitte:
Setting SO_REUSEADDR by default allows server applications to restart and bind to the same address and port without waiting for the OS to fully clean up previous connections. This avoids the issue where we would get "address already in use" after restarting an application after a crash/...

### DIFF
--- a/src/main/cpp/telnetappender.cpp
+++ b/src/main/cpp/telnetappender.cpp
@@ -58,6 +58,7 @@ struct TelnetAppender::TelnetAppenderPriv : public AppenderSkeletonPrivate
 	{ stopAcceptingConnections(); }
 
 	int port;
+	bool reuseAddress = false;
 	ConnectionList connections;
 	LogString encoding;
 	LOG4CXX_NS::helpers::CharsetEncoderPtr encoder;
@@ -112,7 +113,7 @@ void TelnetAppender::activateOptions(Pool& /* p */)
 {
 	if (_priv->serverSocket == NULL)
 	{
-		_priv->serverSocket = ServerSocket::create(_priv->port);
+		_priv->serverSocket = ServerSocket::create(_priv->port, _priv->reuseAddress);
 		_priv->serverSocket->setSoTimeout(1000);
 	}
 
@@ -134,6 +135,10 @@ void TelnetAppender::setOption(const LogString& option,
 	else if (StringHelper::equalsIgnoreCase(option, LOG4CXX_STR("ENCODING"), LOG4CXX_STR("encoding")))
 	{
 		setEncoding(value);
+	}
+	else if (StringHelper::equalsIgnoreCase(option, LOG4CXX_STR("REUSEADDRESS"), LOG4CXX_STR("reuseaddress")))
+	{
+		setReuseAddress(OptionConverter::toBoolean(value, true));
 	}
 	else
 	{
@@ -356,6 +361,11 @@ void TelnetAppender::setMaxConnections(int newValue)
 			--_priv->activeConnections;
 		}
 	}
+}
+
+void TelnetAppender::setReuseAddress(bool reuseAddress)
+{
+	_priv->reuseAddress = reuseAddress;
 }
 
 bool TelnetAppender::requiresLayout() const
--- a/src/main/include/log4cxx/helpers/serversocket.h
+++ b/src/main/include/log4cxx/helpers/serversocket.h
@@ -57,7 +57,12 @@ class LOG4CXX_EXPORT ServerSocket
 		*/
 		void setSoTimeout(int timeout);
 
+#if LOG4CXX_ABI_VERSION <= 15
 		static ServerSocketUniquePtr create(int port);
+		static ServerSocketUniquePtr create(int port, bool reuseAddress);
+#else
+		static ServerSocketUniquePtr create(int port, bool reuseAddress = false);
+#endif
 
 };
 }  // namespace helpers
--- a/src/main/include/log4cxx/private/aprserversocket.h
+++ b/src/main/include/log4cxx/private/aprserversocket.h
@@ -31,7 +31,12 @@ namespace helpers
 class LOG4CXX_EXPORT APRServerSocket : public helpers::ServerSocket
 {
         public:
+#if LOG4CXX_ABI_VERSION <= 15
             APRServerSocket(int port);
+            APRServerSocket(int port, bool reuseAddress);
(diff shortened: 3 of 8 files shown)


swebb2066:
Are you using the telnet appender?

swebb2066:
Given [the uncertainty around this socket option](https://stackoverflow.com/questions/3229860/what-is-the-meaning-of-so-reuseaddr-setsockopt-option-linux), could you also add a AllowQuickReconnect TelnetAppender option?

I think the default value of the option can be True.

This will require adding a setQuickReconnect() method to ServerSocket and a new member variable in APRServerSocketPriv, in addition to TelnetAppender::setOption(0 and TelnetAppender::activateOptions changes.

michieldwitte:
Not much a fan of "QuickReconnect", I would prefer "ReuseAddress" as this will be much more recognizable for anyone who has dealt with sockets.

swebb2066:
> I would prefer "ReuseAddress"

The option name is not super important to me, as any single word is inadequate. 

The most important thing to do is ensure the documentation of the option is clear. An explict description of what it changes should be added the new `set` method of the TelnetAppender class. Also add information in the introduction of TelnetAppender detailing the effects of the option.

src/main/cpp/telnetappender.cpp
```diff
@@ -135,6 +137,10 @@ void TelnetAppender::setOption(const LogString& option,
 	{
 		setEncoding(value);
 	}
+	else if (StringHelper::equalsIgnoreCase(option, LOG4CXX_STR("QUICKRECONNECT"), LOG4CXX_STR("QUICKRECONNECT")))
```
swebb2066, line 140: `else if (StringHelper::equalsIgnoreCase(option, LOG4CXX_STR("QUICKRECONNECT"), LOG4CXX_STR("quickconnect")))`

swebb2066 (commented):
equalIgnoreCase needs an uppercase and lowercase version of the opton name

michieldwitte:
Made some edits as we need to be able to set the option before `apr_socket_bind`.
I opted to pass it to the constructor, let me know if there are other edits required.
Thanks

src/main/include/log4cxx/private/aprserversocket.h
```diff
@@ -31,7 +31,7 @@ namespace helpers
 class LOG4CXX_EXPORT APRServerSocket : public helpers::ServerSocket
 {
         public:
-            APRServerSocket(int port);
+            APRServerSocket(int port, bool reuseAddress = true);
```
rm5248, line 34: This causes an ABI breakage, you need to add a new constructor that takes in two parameters

src/main/include/log4cxx/helpers/serversocket.h
```diff
@@ -57,7 +57,7 @@ class LOG4CXX_EXPORT ServerSocket
 		*/
 		void setSoTimeout(int timeout);
 
-		static ServerSocketUniquePtr create(int port);
+		static ServerSocketUniquePtr create(int port, bool reuseAddress = true);
```
rm5248, line 60: This causes an ABI breakage. You need to add a new method that takes in two parameters.

michieldwitte:
Made the default `false` as to keep consistency with the previous behavior as well.

swebb2066:
Could you also add to the telnetappender unit test please
```
diff --git a/src/test/cpp/net/telnetappendertestcase.cpp b/src/test/cpp/net/telnetappendertestcase.cpp
index b7aebbe5..1e5a7437 100644
--- a/src/test/cpp/net/telnetappendertestcase.cpp
+++ b/src/test/cpp/net/telnetappendertestcase.cpp
@@ -105,6 +105,7 @@ class TelnetAppenderTestCase : public AppenderSkeletonTestCase
                        TelnetAppenderPtr appender(new TelnetAppender());
                        appender->setPort(TEST_PORT);
                        appender->setMaxConnections(1);
+                       appender->setReuseAddress(true);
                        Pool p;
                        appender->activateOptions(p);
                        LoggerPtr root(Logger::getRootLogger());

```

swebb2066:
Also, to fix an existing test case fragility:
```
diff --git a/src/test/cpp/net/socketappendertestcase.cpp b/src/test/cpp/net/socketappendertestcase.cpp
index 2d5accdd..607863f2 100644
--- a/src/test/cpp/net/socketappendertestcase.cpp
+++ b/src/test/cpp/net/socketappendertestcase.cpp
@@ -80,7 +80,7 @@ class SocketAppenderTestCase : public AppenderSkeletonTestCase
                        helpers::ServerSocketUniquePtr serverSocket;
                        try
                        {
-                               serverSocket = helpers::ServerSocket::create(tcpPort);
+                               serverSocket = helpers::ServerSocket::create(tcpPort, true);
                        }
                        catch (std::exception& ex)
                        {
```

swebb2066:
Please also change the .cpp files so building with `LOG4CXX_ABI_VER=16` works
```
diff --git a/src/main/cpp/aprserversocket.cpp b/src/main/cpp/aprserversocket.cpp
index 538a3d2b..b7bd27f5 100644
--- a/src/main/cpp/aprserversocket.cpp
+++ b/src/main/cpp/aprserversocket.cpp
@@ -35,8 +35,10 @@ struct APRServerSocket::APRServerSocketPriv : public ServerSocketPrivate {
        apr_socket_t* socket;
 };

+#if LOG4CXX_ABI_VERSION <= 15
 APRServerSocket::APRServerSocket(int port) :
        APRServerSocket(port, false) {}
+#endif

 APRServerSocket::APRServerSocket(int port, bool reuseAddress) :
        ServerSocket(std::make_unique<APRServerSocketPriv>()){
diff --git a/src/main/cpp/serversocket.cpp b/src/main/cpp/serversocket.cpp
index 705b9181..e696bd1c 100644
--- a/src/main/cpp/serversocket.cpp
+++ b/src/main/cpp/serversocket.cpp
@@ -51,9 +51,11 @@ void ServerSocket::setSoTimeout(int newVal)
        m_priv->timeout = newVal;
 }

+#if LOG4CXX_ABI_VERSION <= 15
 ServerSocketUniquePtr ServerSocket::create(int port){
        return std::make_unique<APRServerSocket>(port);
 }
+#endif

 ServerSocketUniquePtr ServerSocket::create(int port, bool reuseAddress){
        return std::make_unique<APRServerSocket>(port, reuseAddress);

```

PR #497, opened 2025-04-19
Compile and test code designated for the next ABI

swebb2066:


rm5248:
This looks good to me, as I was wondering why we had those `#defines` in the code in the first place.

Should we open up a discussion on what to change in the next ABI version, and a general target date?

swebb2066:
> Should we open up a discussion on what to change in the next ABI version, and a general target date?

I don't think we yet have a clear functional benefit to show to wrrent a new ABI. I expect that will arise as the world evolves.

PR #498, opened 2025-05-08
add ability to set hostname in TelnetAppender

michieldwitte:
Add ability to set hostname in TelnetAppender

### DIFF
--- a/src/main/include/log4cxx/net/telnetappender.h
+++ b/src/main/include/log4cxx/net/telnetappender.h
@@ -117,6 +117,17 @@ class LOG4CXX_EXPORT TelnetAppender : public AppenderSkeleton
 		void setPort(int newValue);
 
 		/**
+		The <b>Hostname</b> on which to accept connections.
+		*/
+		LogString getHostname() const;
+
+		/**
+		Use \c newValue as the Hostname on which to accept connections.
+		By default connections are accepted on any network interface device.
+		*/
+		void setHostname(const LogString& newValue);
+
+		/**
 		The number of allowed concurrent connections.
 
 		\sa setOption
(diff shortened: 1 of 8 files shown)


src/main/include/log4cxx/net/telnetappender.h
```diff
@@ -116,6 +116,16 @@ class LOG4CXX_EXPORT TelnetAppender : public AppenderSkeleton
 		*/
 		void setPort(int newValue);
 
+		/**
+		The <b>Hostname</b> on which to accept connections.
+		*/
+		LogString getHostname() const;
```
swebb2066, line 122: Would 'NodeName' be bettter than 'HostName' given it identifies the local machine interface to be used? 
I believe that would correspond better with [the underlying terminology](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html) and distinguish it from a url used to identify a remote server.

src/main/include/log4cxx/net/telnetappender.h
```diff
@@ -116,6 +116,16 @@ class LOG4CXX_EXPORT TelnetAppender : public AppenderSkeleton
...
+
+		/**
+		Use \c newValue as the Hostname on which to accept connections.
```
swebb2066, line 125: I suggest 'Accept connections on the the \c newValue local interface. By default connections are accepted on any network interface device.'

swebb2066 (commented):
Could you confirm you have been able to test this PR on a machine with multiple network interfaces?

michieldwitte:
> Could you confirm you have been able to test this PR on a machine with multiple network interfaces?

By default it accepts on any network interface, I've validated it using "127.0.0.1" so it only accepts connections on the loopback interface.

src/test/cpp/net/telnetappendertestcase.cpp
```diff
@@ -106,6 +106,7 @@ class TelnetAppenderTestCase : public AppenderSkeletonTestCase
 			appender->setPort(TEST_PORT);
 			appender->setMaxConnections(1);
 			appender->setReuseAddress(true);
+			appender->setHostname("127.0.0.1");
```
swebb2066, line 109: appender->setHostname(;LOG4CXX_STR("127.0.0.1"))

PR #509, opened 2025-07-11
Escape '&' and '"' in html attribute data

swebb2066:
This PR applies [the attibute nodes spec in Canonical XML 2.0](https://www.w3.org/TR/xml-c14n2/#sec-Processing-Model) when modifying attribute strings

### DIFF
--- a/src/main/cpp/htmllayout.cpp
+++ b/src/main/cpp/htmllayout.cpp
@@ -109,18 +109,18 @@ void HTMLLayout::format(LogString& output,
 	if (event->getLevel()->equals(Level::getDebug()))
 	{
 		output.append(LOG4CXX_STR("<font color=\"#339933\">"));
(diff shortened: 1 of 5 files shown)


ppkarwasz:
@swebb2066, what about thread name? What format does thread name in C++ follow?

swebb2066:
The Log4cxx user would have to use [pthread_setname_np](https://linux.die.net/man/3/pthread_getname_np) or [SetThreadDescription](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreaddescription) to establish the thread name.

That is, there are no constraints on the format.

swebb2066:
As HTMLLayout is currentently not consistant with other structured outputs, I have created #514 to make it consistant.

PR #517, opened 2025-08-03
Use doxygen 1.14 to get a more modern looking web site

swebb2066:
The [staging web site](https://logging.staged.apache.org/log4cxx/1.5.0/) is available for review


swebb2066:
If I push the doxygen 1.14 version to asf-site, people who have already visited [the 1.5.0 site](https://logging.apache.org/log4cxx/1.5.0/) will need to clear their cache to get the site to display correctly.

I am thinking that I should leave the 1.5.0 as it is and use the doxygen 1.14 version for the next version.

Thoughts?

rm5248:
> I am thinking that I should leave the 1.5.0 as it is and use the doxygen 1.14 version for the next version.

that sounds like a reasonable plan to me.

PR #520, opened 2025-08-15
Add support for programmatically providing runtime values for configuration option values

swebb2066:
See the example config4.cpp for how this works.

Note: I also removed support for the "log4j.configuration" environment variable as it is not supported by Log4j2 anyway.


rm5248:
> Note: I also removed support for the "log4j.configuration" environment variable as it is not supported by Log4j2 anyway.

Do we want to remove it now, or should we wait for the next major version?  This would be a breaking change.

the reason it is `log4j.configuration` is that it follows the Log4j1 variable name.

I would also recommend that the behavior added in this PR have an option to explicitly disable this new feature via an environment variable.  I'm thinking back to log4shell, which was caused by abusing string replacements.

PR #527, opened 2025-08-28
Improve internal debug logging message consistency in configuration file parsing

swebb2066:
This PR aims to improve the usefulness of Log4cxx internal debug logging


src/main/cpp/domconfigurator.cpp
```diff
@@ -840,13 +846,6 @@ spi::ConfigurationStatus DOMConfigurator::doConfigure
 {
 	m_priv->repository = repository ? repository : LogManager::getLoggerRepository();
 	m_priv->repository->setConfigured(true);
-	if (LogLog::isDebugEnabled())
```
rm5248, line 843: do we print this message out(what file is being used for configuration) anyplace else?  That can be useful to know when you're unsure which file is being used

swebb2066: Yes, a few lines below it outputs either "Loading configuration file [" ... or "Could not read configuration file [" ...

PR #548, opened 2025-09-30
Allow binary-to-text conversion to be moved to a background thread

swebb2066:
This PR introduces to Log4cxx the equivalent of Log4j2's asynchronous loggers

### DIFF
--- a/src/main/include/log4cxx/helpers/asyncbuffer.h
+++ b/src/main/include/log4cxx/helpers/asyncbuffer.h
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


rm5248:
This might break C++11 support, I just took a look at the build:

```
/home/runner/work/logging-log4cxx/logging-log4cxx/main/src/main/include/log4cxx/helpers/asyncbuffer.h:72:25: warning: lambda capture initializers only available with ‘-std=c++14’ or ‘-std=gnu++14’ [-Wc++14-extensions]
   72 |                 append([value = std::move(rvalue)](LogCharMessageBuffer& msgBuf)
      |                         ^~~~~
```

The build does still pass though - should we turn on `Werror` for warnings like this in order to fail the build?

src/test/cpp/asyncappendertestcase.cpp
rm5248, line 1: It looks like this changes the tests, since the macro changes from `LOG4CXX_DEBUG` to `LOG4CXX_DEBUG_ASYNC`.  Do we need to have two tests, one that uses the old macro and one that uses the new?  I would be worried about breaking existing code if we don't keep the current tests as-is.

src/main/include/log4cxx/helpers/asyncbuffer.h
```diff
@@ -0,0 +1,273 @@
...
+	<< " at " << obj->getBoundingBox().getCenter()
+	<< " +/- " << obj->getBoundingBox().getHalfSize()
+	);
+~~~
+
+@param logger the logger that has the enabled status.
+@param message a valid r-value expression of an <code>operator<<(std::ostream&. ...)</code> overload.
+
+*/
+#define LOG4CXX_DEBUG_ASYNC(logger, message) do { \
```
rm5248, line 134: do we need to add the `_FMT` family of macros to these as well, or does that not make sense?  fmt is definitely easier to work with

PR #561, opened 2025-11-02
Show the performance results in graphs

rm5248:
Show performance results in a graphical way

This shows the performance results in a clearer manner than just the table, and makes it more obvious the relative difference between the different scenarios.


swebb2066:
The Iteration column is not the data, just an implementation detail.

The actual data is the average (elapsed) time and CPU

swebb2066:
Can you script the graph generation?

PR #585, opened 2026-01-27
Fix: Integer Overflow in OptionConverter & Enable >2GB logs on Windows

OxBat:
### Summary
I identified a platform-dependent integer overflow in `OptionConverter::toFileSize`.
The library previously used `long` for file size calculations. On Windows (MSVC), `long` is strictly 32-bit (LLP64 data model), whereas it is 64-bit on Linux (LP64).

**The Issue:**
Configuring `MaxFileSize="3GB"` on Windows triggers a signed integer overflow:
`3 * 1024^3` = `3,221,225,472` -> wraps to `-1,073,741,824` (negative).
This negative value is passed to RollingFileAppender, causing logic errors (DoS via infinite rotation or rotation failure).

**The Fix:**
I updated the API signature in `optionconverter.h` and the implementation in `optionconverter.cpp` to use `long long` (guaranteed 64-bit on all platforms) and leveraged `StringHelper::toInt64`.

### Impact
1. **Fixes the Crash/DoS:** Prevents negative file size limits.
2. **Enables Feature:** Windows users can finally use log files larger than 2GB, ensuring parity with Linux.

### DIFF
--- a/src/main/include/log4cxx/helpers/optionconverter.h
+++ b/src/main/include/log4cxx/helpers/optionconverter.h
@@ -20,6 +20,7 @@
 
 #include <log4cxx/logstring.h>
 #include <log4cxx/helpers/object.h>
+#include <cstdint>
 
 namespace LOG4CXX_NS
 {
@@ -79,7 +80,7 @@ class LOG4CXX_EXPORT OptionConverter
 		 * The numeric equivalent of \c value if it is not empty, otherwise \c defaultValue.
 		 * Supports 64-bit values for file sizes > 2GB.
 		 */
-		static long long toFileSize64(const LogString& value, long long defaultValue);
+		static int64_t toFileSize64(const LogString& value, int64_t defaultValue);
 		/**
 		The Level indicated by \c value if recognised otherwise \c defaultValue.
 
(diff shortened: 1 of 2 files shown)


src/main/include/log4cxx/helpers/optionconverter.h
```diff
@@ -72,7 +72,16 @@ class LOG4CXX_EXPORT OptionConverter
 		 converts the provided number respectively to kilobytes, megabytes
 		 and gigabytes. For example, the value "10KB" will be interpreted as 10240.
 		*/
+		/**
+		 * @deprecated Use toFileSize64 instead.
```
rm5248: this should use the newer `[[deprecated]]` annotation like is used in other places(example: https://github.com/apache/logging-log4cxx/blob/28a443edaf226b37996bc653aed9b4d666785458/src/main/include/log4cxx/helpers/class.h#L37)

src/main/include/log4cxx/helpers/optionconverter.h
```diff
@@ -72,7 +72,16 @@ class LOG4CXX_EXPORT OptionConverter
...
+		 */
 		static long toFileSize(const LogString& value, long defaultValue);
+
+		/**
+		 * The numeric equivalent of \c value if it is not empty, otherwise \c defaultValue.
+		 * Supports 64-bit values for file sizes > 2GB.
+		 */
+		static long long toFileSize64(const LogString& value, long long defaultValue);
```
rm5248: @swebb2066 any thoughts on `long long` vs. `int64_t`?  The latter is used in a few parts of the API, but they should both be equivalent.

swebb2066: > any thoughts on long long vs. int64_t


I prefer int64_t for its explicit size and the single word

swebb2066:
#654 provides a better solution

PR #635, opened 2026-04-23
Use length-bounded comparison in StringHelper::equalsIgnoreCase

jmestwa-coder:
## Summary

Refactor `StringHelper::equalsIgnoreCase` to use explicit length-bounded comparison instead of pointer traversal.

---

## What changed

* Use `std::char_traits<logchar>::length(upper)` to determine length
* Check `s1.length()` before comparison
* Replace pointer increments with index-based access
* Preserve embedded NUL handling

---

## Why

* Removes implicit pointer traversal
* Makes bounds explicit and easier to reason about
* Keeps behavior unchanged

---

## Behavior

No behavior change intended.

---

## Testing

Added tests for:

* case-insensitive match
* length mismatch
* embedded NUL

All tests pass.

---

## Notes

All current call sites pass paired uppercase/lowercase literals of identical length, which this change relies on.

### DIFF
--- a/src/main/cpp/stringhelper.cpp
+++ b/src/main/cpp/stringhelper.cpp
@@ -24,24 +24,36 @@
 #include <iterator>
 #include <algorithm>
 #include <cctype>
+#include <string>
 
 using namespace LOG4CXX_NS;
 using namespace LOG4CXX_NS::helpers;
 
 bool StringHelper::equalsIgnoreCase(const LogString& s1, const logchar* upper, const logchar* lower)
 {
+	const size_t len_upper = std::char_traits<logchar>::length(upper);
+	if (s1.length() != len_upper)
+	{
+		return false;
+	}
+
+	size_t i = 0;
 	for (const auto& item : s1)
 	{
-		if (0 == item || // OSS-Fuzz makes strings with embedded NUL characters
-			(item != *upper && item != *lower))
+		if (item == 0) // OSS-Fuzz makes strings with embedded NUL characters
 		{
 			return false;
 		}
-		++upper;
-		++lower;
+
+		if (item != upper[i] && item != lower[i])
+		{
+			return false;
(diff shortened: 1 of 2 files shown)


src/main/cpp/stringhelper.cpp
```diff
@@ -24,24 +24,33 @@
...
 #include <cctype>
+#include <string>
 
 using namespace LOG4CXX_NS;
 using namespace LOG4CXX_NS::helpers;
 
 bool StringHelper::equalsIgnoreCase(const LogString& s1, const logchar* upper, const logchar* lower)
 {
-	for (const auto& item : s1)
+	const size_t len_upper = std::char_traits<logchar>::length(upper);
```
swebb2066, line 34: I cannot see that this is an improvement as it goes against [this C++ guideline ]( rule ES.71 suggests "Prefer range-for over indexed for").

I would also point out that the indexing is not inherently safer than the pointer increments in this case.

swebb2066 (commented):
Please tell me while this is an improvement.

src/main/cpp/stringhelper.cpp
jmestwa-coder: Thanks for the feedback. I've updated the implementation to retain the range-based loop while keeping the explicit length check.

swebb2066: I still do not see how indexing is better than pointer increments.
Also, I believe the length check ( which is redundent) is less clear than the in-loop test for a 0 item.

jmestwa-coder: Thanks for the clarification. That makes sense — the current implementation is already clear and aligns with the preferred style. I’ll close this PR.

swebb2066: Thanks

PR #636, opened 2026-04-24
Use the argument-free activateOptions() syntax

swebb2066:
This PR changes all test code that calls activateOptions() and adds a test that demonstrates when an ABI-15-style user-defined appender will work and when it will not (it will generate a compilation error).


src/main/include/log4cxx/spi/optionhandler.h
```diff
@@ -43,19 +43,23 @@ class LOG4CXX_EXPORT OptionHandler : public virtual helpers::Object
...
-		related options that remain ambigous until all are set.
+		related options that remain ambiguous until all are set.
 
 		<p>For example, the FileAppender has
 		the <code>File</code> and <b>Append</b> options both of
-		which are ambigous until the other is also set.
+		which are ambiguous until the other is also set.
 		*/
 #if LOG4CXX_ABI_VERSION <= 15
+#define LOG4CXX_ACTIVATE_OPTIONS_FORMAL_PARAMETERS helpers::Pool& p
```
rm5248, line 53: is the intention to remove these macros for the next ABI version?  It's very confusing to see a method with a macro in the parameter list that may or may not be defined.

swebb2066: Yes, when switching to the new ABI, conditional sections can be resolved using [my Boost Wave-based program](https://github.com/swebb2066/cpp-resolve) or [unifdef](https://linux.die.net/man/1/unifdef) and the ABI version-dependant macros (LOG4CXX_ACTIVATE_OPTIONS_FORMAL_PARAMETERS, LOG4CXX_ACTIVATE_OPTIONS_PARAMETER and LOG4CXX_16_VIRTUAL_SPECIFIER) substituted.

PR #652, opened 2026-05-08
Simplify socket appender sub-classing by enhancing SocketAppenderSkeleton

swebb2066:
This PR moves documentation and functionality from XMLSocketAppender to its base class, SocketAppenderSkeleton.

A different network implementation is still allowed, as the Socket class provides a factory method that provides a Socket instance.

This change will allow, for example, the addition of a socket appender without a default layout.

### DIFF
--- a/src/main/cpp/visualstudio.natvis
+++ b/src/main/cpp/visualstudio.natvis
@@ -59,7 +59,7 @@
 	<Type Name="log4cxx::net::XMLSocketAppender">
 		<Expand>
 			<Item Name="m_priv">
-				static_cast&lt;XMLSocketAppenderPriv*&gt;(m_priv._Mypair._Myval2)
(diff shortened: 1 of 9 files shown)


src/main/cpp/visualstudio.natvis
rm5248, line 1: I haven't noticed this file before, I'm guessing that it's something for Visual Studio to use?  Should it be in the repo in the first place?

swebb2066: > Should it be in the repo in the first place?

Absolutely. It allows developers to immediately see the values of all the attributes of the derived (private) structs in the Visual Studio debugger. It is included in the CMakeLists files for WIN32.
```
if(WIN32)
  target_sources(log4cxx
    PRIVATE visualstudio.natvis
  )
endif(WIN32)

```

PR #658, opened 2026-05-10
#643 Log complete messages with embedded nul characters in console writers

### LINKED ISSUE #643: Logging text with embedded nul bytes truncates log message
Logging a string that contains an embedded nul byte to the console truncates the remaining message:
```cpp
LOG4CXX_INFO(kLogger, "Hello\0World!");
LOG4CXX_INFO(kLogger, "Hello World!");
// [2026-05-03T13:05:42] root INFO - Hello[2026-05-03T13:05:42] root INFO - Hello World
```
Ideally the nul byte should either be skipped or written out as an `\0` escaped string.
I belive this is due to [systemoutwriter.cpp](https://github.com/apache/logging-log4cxx/blob/master/src/main/cpp/systemoutwriter.cpp#L77-L78) which uses `fputs` rather than a print method thats aware of nul bytes

apache-hb:
Resolves #643 
Writes to stdout and stderr now go through fwrite rather than fputs. This stops embedded nul characters in log messages from truncating the rest of the message. Added a regression test `testWriteEmbeddedNullCharacters` to go along with the changes.

### DIFF
--- a/src/test/cpp/helpers/consolewritertestcase.cpp
+++ b/src/test/cpp/helpers/consolewritertestcase.cpp
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "../logunit.h"
+
+#include <log4cxx/private/consolewriter_priv.h>
+#include <log4cxx/helpers/pool.h>
+#include <apr_file_io.h>
+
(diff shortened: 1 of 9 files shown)


swebb2066:
I think moving the duplicated `isWide()` implementations to a single implementation in `consolewrite.cpp` would improve the code.

Note that both implementations are currently `private`, so can be removed from the class interface without breaking the ABI.

src/test/cpp/helpers/consolewritertestcase.cpp
```diff
@@ -0,0 +1,75 @@
...
+		std::string path = tmpdir;
+		path += "/log4cxx.test.log";
+
+		remove(path.c_str());
+
+		std::string message{"Hello\0World!", 12};
+
+		{
+			FILE *fp = fopen(path.c_str(), "wb");
+			size_t written = helpers::writeToConsole(false, message, fp);
```
swebb2066: This should be:
```
LOG4CXX_DECODE_CHAR(lsMessage, message);
size_t written = helpers::writeToConsole(lsMessage, fp);
```

src/test/cpp/helpers/consolewritertestcase.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "../logunit.h"
+
+#include <log4cxx/private/consolewriter_priv.h>
+#include <log4cxx/helpers/pool.h>
```
swebb2066, line 21: Just need the following 
`#include <log4cxx/helpers/transcoder.h>`

PR #665, opened 2026-05-13
Reject trailing characters in StringHelper numeric parsing

jmestwa-coder:
## Summary

Improve `StringHelper::toInt` and `StringHelper::toInt64` to reject malformed numeric input with trailing non-whitespace characters instead of silently accepting partial parses.

Examples of previously accepted input:
- `"123abc"` → parsed as `123`
- `"514udp"` → parsed as `514`

This change now requires the full numeric string to be consumed (excluding trailing whitespace).

## Changes

- Added full-consumption validation after `std::stoi` / `std::stoll`
- Preserved support for surrounding whitespace and signed values
- Switched to `LOG4CXX_ENCODE_CHAR` for consistent cross-platform `LogString` handling
- Added regression tests covering:
  - trailing characters
  - empty input
  - whitespace-only input
  - embedded whitespace
  - signed numbers

## Rationale

`OptionConverter::toInt` already rejects trailing characters using full-string validation. This change aligns `StringHelper` numeric parsing behavior with the existing strict parsing semantics used elsewhere in the project and avoids silently accepting malformed numeric configuration values.

### DIFF
--- a/src/main/cpp/stringhelper.cpp
+++ b/src/main/cpp/stringhelper.cpp
@@ -24,9 +24,26 @@
 #include <iterator>
 #include <algorithm>
 #include <cctype>
+#include <stdexcept>
 
 using namespace LOG4CXX_NS;
 using namespace LOG4CXX_NS::helpers;
+
+namespace
+{
+void checkFullyParsed(const std::string& value, std::size_t pos)
+{
+	while (pos < value.size() && std::isspace(static_cast<unsigned char>(value[pos])))
+	{
+		++pos;
+	}
+
+	if (pos != value.size())
+	{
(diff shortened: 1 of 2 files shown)


swebb2066:
As the failing tests show, this PR breaks existing code that expects trailing non-digits to terminate an integer.

I do not believe this PR adds any value to the code base, especially given that `OptionConverter::toInt` is available for cases where trailing characters are an error.

jmestwa-coder:
@swebb2066 , I was treating the behavior difference as an inconsistency, but I understand now that the permissive parsing in StringHelper::toInt is intentional and relied upon by existing code paths.

PR #671, opened 2026-05-14
Prevent rollover failure with short file name patterns

jmestwa-coder:
## Summary
Prevent rollover failures when using short file name patterns in `FixedWindowRollingPolicy`.

## What Changed
- Replaced manual suffix offset comparisons with `StringHelper::endsWith()`
- Removed unsafe `length() - N` arithmetic for `.gz` and `.zip` suffix checks
- Added a regression test covering short patterns such as `R%i`

## Problem
Short valid patterns like:

```text
R%i
```

can generate file names shorter than the suffix length being checked.

The previous implementation performed:

```cpp
lowFilename.compare(lowFilename.length() - 3, ...)
```

which could produce an invalid string position and cause rollover to fail with:

```text
log4cxx: invalid string position
```

## Regression Test
Added `testShortPatternRegression` to verify:
- rollover fails before the fix
- rollover succeeds after the fix
- the rolled file is created successfully

## Validation
- Verified failing behavior on the original implementation
- Verified passing behavior after the fix
- Re-ran the regression test multiple times to confirm deterministic behavior

### DIFF
--- a/src/main/cpp/fixedwindowrollingpolicy.cpp
+++ b/src/main/cpp/fixedwindowrollingpolicy.cpp
@@ -265,11 +265,11 @@ bool FixedWindowRollingPolicy::purge(int lowIndex, int highIndex, Pool& p) const
 
 	LogString lowFilename(buf);
 
-	if (lowFilename.compare(lowFilename.length() - 3, 3, LOG4CXX_STR(".gz")) == 0)
+	if (StringHelper::endsWith(lowFilename, LOG4CXX_STR(".gz")))
 	{
 		suffixLength = 3;
 	}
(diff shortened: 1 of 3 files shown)


rm5248:
@jmestwa-coder  since you have opened several pull requests, if you have not already please sign and send in your [ICLA](https://www.apache.org/licenses/contributor-agreements.html) and post back here once you have done so

jmestwa-coder:
> @jmestwa-coder since you have opened several pull requests, if you have not already please sign and send in your [ICLA](https://www.apache.org/licenses/contributor-agreements.html) and post back here once you have done so

Done

PR #677, opened 2026-05-16
Adds `AGENTS.md` file

ppkarwasz:
Adds an `AGENTS.md` file to guide LLMs to our common security/threat model.

In the long term we should probably have specific threat models for each project, since some aspects obviously differ between the three implementations (e.g. memory safety), but we'll have to figure out a way to keep a common threat model and only add the aspects that differ to the various projects.


FreeAndNil (approved):
LGTM

PR #682, opened 2026-05-19
SyslogAppender packet splitting for large chunk counts

jmestwa-coder:
## Summary
Fix SyslogAppender packet splitting when large chunk counts cause the `(x/y)` suffix to exceed the reserved suffix space.

## What changed
- Replaced iterator-based splitting with bounded `size_t` index arithmetic.
- Added suffix-aware chunk sizing to keep emitted payloads within `MaxMessageLength`.
- Added explicit handling for cases where the full `(x/y)` suffix cannot fit alongside payload data.
- Replaced fixed `apr_snprintf` suffix formatting with `StringHelper::toString`.
- Added defensive guards around packet reservation for extremely large message counts.

## Behavior
- Normal packet splitting behavior is preserved for standard message sizes.
- When the suffix cannot fit within `MaxMessageLength`, the appender now omits the suffix instead of emitting oversized packets.
- A warning is logged when suffix omission occurs.

## Tests
Added regression coverage for:
- one-byte remainder splitting boundaries
- large chunk-count suffix growth
- payload length invariant enforcement
- impossible suffix-fit scenarios

## Result
Ensures emitted SyslogAppender payloads remain bounded by `MaxMessageLength` across large message and packet-count edge cases.

### DIFF
--- a/src/main/cpp/syslogappender.cpp
+++ b/src/main/cpp/syslogappender.cpp
@@ -55,7 +55,7 @@ namespace net
 {
 namespace detail
 {
-std::vector<LogString> splitSyslogPackets(const LogString& msg, size_t maxMessageLength)
+LOG4CXX_EXPORT std::vector<LogString> splitSyslogPackets(const LogString& msg, size_t maxMessageLength)
 {
 	std::vector<LogString> packets;
 	auto digitCount = [](size_t value) {
--- a/src/main/cpp/syslogappender.cpp
+++ b/src/main/cpp/syslogappender.cpp
@@ -26,14 +26,150 @@
 #if !defined(LOG4CXX)
 	#define LOG4CXX 1
 #endif
-#include <apr_strings.h>
 #include <log4cxx/private/syslogappender_priv.h>
+#include <algorithm>
 
 #define LOG_UNDEF -1
 
 using namespace LOG4CXX_NS;
 using namespace LOG4CXX_NS::helpers;
 using namespace LOG4CXX_NS::net;
+
+namespace
+{
+	constexpr size_t SYSLOG_PACKET_SUFFIX_RESERVED_CHARS = 12;
+
+	void appendPacketSuffix(LogString& item, size_t current, size_t total)
+	{
+		item.append(LOG4CXX_STR("("));
+		StringHelper::toString(current, item);
+		item.append(LOG4CXX_STR("/"));
+		StringHelper::toString(total, item);
+		item.append(LOG4CXX_STR(")"));
+	}
+}
+
+namespace LOG4CXX_NS
+{
+namespace net
+{
+namespace detail
+{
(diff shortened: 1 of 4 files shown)


src/main/cpp/syslogwriter.cpp
```diff
@@ -64,6 +64,16 @@ SyslogWriter::~SyslogWriter(){}
 
 void SyslogWriter::write(const LogString& source)
 {
+#if defined(UNIT_TEST)
```
rm5248: how would this block ever get activated?

jmestwa-coder: `UNIT_TEST` is defined in `syslogappendertestcase.cpp`, so this path is only enabled for those tests.

The goal was to make the packet-size assertions deterministic without depending on UDP/network behavior. If there’s a preferred way to structure this kind of test in log4cxx, I’m happy to adjust it.

rm5248: It would never get compiled with `UNIT_TEST` defined though, you would have to compile the library with that macro defined.

Instead, it would make more sense to make a helper/utility function that can take a string of arbitrary length  and split it up into smaller strings of a given max length.  This function would then get called by the `SyslogAppender` to split the log message up.  This function can then be properly tested on its own.

jmestwa-coder: Thanks for the suggestion .

I reworked the patch to move the splitting logic into a dedicated helper that `SyslogAppender` calls directly, removed the `UNIT_TEST` interception path, and updated the tests to validate the helper behavior directly instead of capturing network writes.

If there’s still a simpler structure you’d prefer here, I’m happy to adjust it further.

jmestwa-coder:
pushed a fix for the CI build failures: the helper was declared outside the net namespace and the test needed LOG4CXX_TEST defined to use the private header.

src/main/cpp/syslogappender.cpp
```diff
@@ -26,15 +26,151 @@
...
+	}
+}
+
+namespace LOG4CXX_NS
+{
+namespace net
+{
+namespace detail
+{
+std::vector<LogString> splitSyslogPackets(const LogString& msg, size_t maxMessageLength)
```
swebb2066: `LOG4CXX_EXPORT std::vector<LogString> splitSyslogPackets(const LogString& msg, size_t maxMessageLength)`

swebb2066 (commented):
An alternative is adding a public static method SysLogAppender::asPackets() to SysLogAppender.

src/main/cpp/syslogappender.cpp
jmestwa-coder: done, added LOG4CXX_EXPORT to the declaration and definition. that was the cause of the link failures on the shared-library builds, and the test now links and passes locally. happy to move it to a public static SyslogAppender::asPackets() instead if you'd rather expose it on the class.

PR #686, opened 2026-05-23
JSONLayout escaping for malformed UTF-8 input

jmestwa-coder:
## Summary

Fix JSONLayout escaping for malformed UTF-8 input.

Previously, malformed decoded input could bypass replacement escaping and preserve the original malformed bytes in emitted JSON output. This happened because the replacement character (`U+FFFD`) incorrectly followed the printable fast path.

This change ensures malformed decoded input and invalid Unicode scalar values are consistently emitted through the JSON escaping path as `\ufffd`.

## Changes

- Add `escapeReplacement` handling in `JSONLayout::appendItem()`
- Ensure replacement characters bypass the printable fast path
- Preserve existing behavior for valid UTF-8 input
- Add regression coverage for malformed UTF-8 escaping behavior

## Regression Test

Added a focused regression test for malformed UTF-8 input to verify malformed bytes no longer survive into JSON output and are emitted as escaped replacement characters instead.

### DIFF
--- a/src/main/cpp/jsonlayout.cpp
+++ b/src/main/cpp/jsonlayout.cpp
@@ -223,18 +223,21 @@ void JSONLayout::appendItem(const LogString& input, LogString& buf)
 	{
 		auto lastCodePoint = nextCodePoint;
 		auto ch = Transcoder::decode(input, nextCodePoint);
+		bool escapeReplacement = false;
 		if (nextCodePoint == lastCodePoint) // failed to decode input?
 		{
 			nextCodePoint = input.end();
 			ch = 0xFFFD; // The Unicode replacement character
+			escapeReplacement = true;
 		}
 		else if ((0xD800 <= ch && ch <= 0xDFFF) || 0x10FFFF < ch)
 		{
 			ch = 0xFFFD; // The Unicode replacement character
+			escapeReplacement = true;
 		}
 		else if (0x22 == ch || 0x5c == ch) // double quote or backslash?
 			;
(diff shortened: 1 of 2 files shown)


src/main/cpp/jsonlayout.cpp
```diff
@@ -223,18 +223,21 @@ void JSONLayout::appendItem(const LogString& input, LogString& buf)
...
 		}
 		else if ((0xD800 <= ch && ch <= 0xDFFF) || 0x10FFFF < ch)
 		{
 			ch = 0xFFFD; // The Unicode replacement character
+			escapeReplacement = true;
 		}
 		else if (0x22 == ch || 0x5c == ch) // double quote or backslash?
 			;
-		else if (0x20 <= ch) // not a control character?
+		else if (!escapeReplacement && 0x20 <= ch) // not a control character?
```
swebb2066, line 240: This change does nothing as `escapeReplacement` is always `false` in this conditional expression.

swebb2066 (commented):
I suggest you just add the unit test and leave `jsonlayout` unchanged.

src/main/cpp/jsonlayout.cpp
jmestwa-coder: Thanks for the review and for pointing this out.

I re-validated the runtime behavior against the malformed UTF-8 testcase and confirmed the change does not actually alter the emitted output as intended. The new condition is effectively a no-op in the current control flow.

Given that, I’m going to withdraw this patch rather than push a speculative follow-up change without a demonstrated failing case.

PR #696, opened 2026-06-01
Reject negative length and offset values in DatagramPacket

jmestwa-coder:
## Summary

Reject negative length and offset values when creating or modifying `DatagramPacket` instances.

## Changes

* Add centralized validation for `DatagramPacket` length and offset arguments.
* Reject negative length values in all constructors and mutators.
* Reject negative offset values in all constructors and mutators.
* Prevent invalid packet state from being stored and propagated.
* Add regression tests covering negative constructor and mutation inputs.
* Register the new test target in the helper test suite.

## Rationale

`DatagramPacket` currently accepts negative length and offset values and stores them unchanged. This change enforces basic argument invariants at the `DatagramPacket` boundary and ensures invalid values are rejected before they can propagate through socket operations.

## Testing

* Added constructor validation tests for negative length and offset values.
* Added mutation validation tests for negative `setLength()` and `setData()` inputs.
* Verified focused regression coverage for all supported negative-input paths.

### DIFF
--- a/src/main/cpp/datagrampacket.cpp
+++ b/src/main/cpp/datagrampacket.cpp
@@ -20,25 +20,34 @@
 
 using namespace LOG4CXX_NS::helpers;
 
+namespace
+{
+#if LOG4CXX_ABI_VERSION <= 15
+using PacketSizeType = int;
+#else
+using PacketSizeType = std::size_t;
+#endif
+}
+
 struct DatagramPacket::DatagramPacketPriv
 {
-	DatagramPacketPriv(void* buf1, int length1)
+	DatagramPacketPriv(void* buf1, PacketSizeType length1)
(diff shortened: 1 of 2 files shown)


swebb2066:
I would like to change the interface in the next ABI version to use `size_t` instead of `int`.

The test case would then be unnecessary.

Would you like to introduce the interface changes for the next ABI in this PR?

jmestwa-coder:
Using `size_t` for the length and offset parameters does seem like a cleaner way to enforce the invariant at the API level.

I'm happy to rework the PR around the interface change if that's the preferred direction. Is there a specific branch or upcoming ABI version where you would like these changes to target?

rm5248:
master branch is fine, you need to use the `LOG4CXX_ABI_VERSION` macro

PR #698, opened 2026-06-01
Honor DatagramPacket offsets and update received packet length

jmestwa-coder:
## Summary

This change fixes DatagramPacket handling in the UDP socket implementation by honoring packet offsets during send/receive operations and updating the packet length after a successful receive.

## Changes

* Honor `DatagramPacket::getOffset()` when sending UDP packets.
* Honor `DatagramPacket::getOffset()` when receiving UDP packets.
* Update the packet length to the actual number of bytes received.
* Add centralized validation for invalid packet states:

  * Reject negative offsets.
  * Reject negative lengths.
  * Reject null buffers for non-empty packets.

### DIFF
--- a/src/main/cpp/aprdatagramsocket.cpp
+++ b/src/main/cpp/aprdatagramsocket.cpp
@@ -97,13 +97,20 @@ void APRDatagramSocket::receive(DatagramPacketPtr& p)
 
 	// receive the datagram packet
 	apr_size_t len = p->getLength();
+	auto data = static_cast<char*>(p->getData());
+	if (data != nullptr)
+	{
+		data += p->getOffset();
+	}
 	status = apr_socket_recvfrom(addr, _priv->socket, 0,
-			(char*)p->getData(), &len);
+			data, &len);
 
 	if (status != APR_SUCCESS)
 	{
 		throw IOException(status);
 	}
+
+	p->setLength(static_cast<int>(len));
 }
 
 void APRDatagramSocket::send(DatagramPacketPtr& p)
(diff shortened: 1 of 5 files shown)


rm5248:
It this really needed?  Some thoughts:

1. Why would we be receiving packets?  I don't recall any code that actually receives packets, but perhaps we do someplace
2. Is there a need to honor the offset?  Is the offset even required?  At least when we're sending data it should just be a full pcaket or not, I don't know why we would want to send part of a packet.

jmestwa-coder:
My understanding was that `DatagramPacket` exposes offset-aware constructors and stores both an offset and length, which led me to interpret the offset as part of the packet contract rather than unused state.

The tests I added demonstrate that the current implementation ignores the stored offset during send/receive and does not update the packet length after receive.

That said, if the intended design is that offsets are not meant to influence socket operations, or if receive paths are effectively unused in log4cxx, I'm happy to narrow the change or revisit the approach.

jmestwa-coder:
any update?

swebb2066:
I believe this PR has been made obsolete by your other PR #696.

Can you address conflicts if this is not the case.

jmestwa-coder:
Agreed, #696 covers this. The size_t switch there makes the negative offset/length checks in this PR redundant, and it conflicts in datagrampacket.cpp for the same reason. The only remaining piece is the socket-side offset handling, which is the part rm5248 wasn't sold on anyway, so it's not worth rebasing on its own. Closing this one and keeping #696.

PR #707, opened 2026-06-04
Fix application hang on SocketAppender shutdown when offline

sahvx655-wq:
Fixes an application hang during SocketAppender shutdown when the destination log server is offline.
ThreadUtility::priv_data::doPeriodicTasks() executed periodic task callbacks while holding job_mutex. If a reconnect attempt blocked on network I/O, the mutex remained held, preventing removePeriodicTask() and shutdown-related operations from acquiring the lock and causing shutdown to stall.
This change executes task callbacks outside of job_mutex, eliminating lock contention between long-running callbacks and task removal/shutdown paths.
Added testNoDeadlockDuringTaskExecution to verify that removePeriodicTask() completes successfully while a slow periodic task is still executing.


swebb2066:
> Added testNoDeadlockDuringTaskExecution to verify that removePeriodicTask() completes successfully while a slow periodic task is still executing.

Could you push this test case please?

sahvx655-wq:
Sorry, I'd left it out of the commit. Pushed it now as testNoDeadlockDuringTaskExecution in threadutilitytestcase.cpp.

It adds a periodic task whose callback blocks (standing in for a reconnect stuck on network I/O), waits for the callback to start, then times removePeriodicTask() while that callback is still running. Without the change removePeriodicTask() blocks on job_mutex until the callback returns; with it the removal comes back in a few ms. Runs green here alongside the existing cases.

src/main/cpp/threadutility.cpp
```diff
@@ -369,39 +369,86 @@ void ThreadUtility::priv_data::doPeriodicTasks()
...
+					);
+
+				if (pItem != this->jobs.end())
+				{
+					if (success)
 					{
-						LogLog::warn(item.name, ex);
-						++item.errorCount;
+						pItem->nextRun = std::chrono::system_clock::now() + taskDelay;
+						pItem->errorCount = 0;
```
swebb2066, line 435: Update pItem->nextRun unconditionally. That will create some time between retries when the task fails.

swebb2066 (commented):
Please state the intent of every block in a comment before the block as you have dome for some blocks. It really helps maintainers to be able to see intent.

src/main/cpp/threadutility.cpp
```diff
@@ -369,39 +369,86 @@ void ThreadUtility::priv_data::doPeriodicTasks()
...
+			bool foundTask = false;
+
 			{
-				if (this->terminated.load())
-					return;
-				if (item.removed)
-					;
-				else if (item.nextRun <= currentTime)
+				std::lock_guard<std::recursive_mutex> lock(this->job_mutex);
+				for (auto& item : this->jobs)
```
swebb2066: I think this would be clearer using std::find_if

swebb2066 (commented):
Using std::find_if would improve consistency.

swebb2066:
Also, please merge the master branch to include the updated CI actions - the 10GB cache only allows for 1 Windows-qt build

src/main/cpp/threadutility.cpp
sahvx655-wq: Done. nextRun now gets pushed out on both paths, so a task that keeps failing (a reconnect that can't reach the server) waits taskDelay before the next attempt instead of spinning. errorCount still only resets on success.

src/main/cpp/threadutility.cpp
sahvx655-wq: Switched the selection loop over to std::find_if as well, so it matches the reschedule and cleanup blocks. The terminated check now sits just before we take the lock.

sahvx655-wq:
Pushed an update covering the rest. I've added an intent comment before each block in doPeriodicTasks() now, including the select, execute, reschedule and wait blocks, so the flow reads top to bottom.

Also merged master in, so the updated Windows CI actions are picked up and the qt build should fit inside the cache budget.

PR #715, opened 2026-06-08
Fix ignored gzip exit status in GZCompressAction

sahvx655-wq:
GZCompressAction::execute currently waits for the spawned gzip process using apr_proc_wait, but discards the child process exit status by passing NULL for the exit code parameter.

As a result, failures from the compression process are not explicitly detected before execution continues. This behavior differs from ZipCompressAction::execute, which retrieves and validates the child process exit status.

This patch updates GZCompressAction::execute to capture the exit status returned by apr_proc_wait and validate it before proceeding, aligning its behavior with the existing implementation in ZipCompressAction.

Impact
Improves error handling consistency for compression actions and ensures that gzip process failures are not silently ignored.
Proposed Changes
Retrieve the child process exit status in GZCompressAction::execute.
Validate the returned exit status before continuing.
Align gzip compression handling with the existing ZipCompressAction implementation.

### DIFF
--- a/src/main/cpp/gzcompressaction.cpp
+++ b/src/main/cpp/gzcompressaction.cpp
@@ -153,12 +153,18 @@ bool GZCompressAction::execute( LOG4CXX_EXECUTE_ACTION_FORMAL_PARAMETERS ) const
 			return true;
 		}
 
-		apr_proc_wait(&pid, NULL, NULL, APR_WAIT);
+		int exitCode = 0;
+		apr_proc_wait(&pid, &exitCode, NULL, APR_WAIT);
 		stat = apr_file_close(child_out);
 
 		if (stat != APR_SUCCESS)
 		{
 			throw IOException(stat);
+		}
+
+		if (exitCode != APR_SUCCESS)
+		{
+			throw IOException(exitCode);
 		}
 
 		priv->destination.setAutoDelete(false);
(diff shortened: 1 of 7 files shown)


src/main/cpp/gzcompressaction.cpp
```diff
@@ -153,14 +153,20 @@ bool GZCompressAction::execute( LOG4CXX_EXECUTE_ACTION_FORMAL_PARAMETERS ) const
...
 		stat = apr_file_close(child_out);
 
 		if (stat != APR_SUCCESS)
 		{
 			throw IOException(stat);
 		}
 
+		if (exitCode != APR_SUCCESS)
+		{
+			throw IOException(exitCode);
```
swebb2066: The `IOException` parameter is passed to `apr_strerror` to produce a user intelligable message. `apr_strerror` does not expect process an exit code. `zipcompressaction.cpp` is not a good exemplar for this.

I suggest you add an Exception type, say SubProcessFailure, whose constructor takes a process name, exit code and reason `apr_exit_why_e`

```
			apr_proc_wait(&pid[i], &exitCode, &reason, APR_WAIT);
			if (exitCode != 0)
			{
               throw SubProcessFailure("gzip", exitCode, reason);
            }

```
where SubProcessFailure::makeMessage may do something like:
```
              {
				std::string msg = processName;
				msg += " exit code: ";
				msg += std::to_string(exitCode);
                if (APR_PROC_SIGNAL  == reason || APR_PROC_SIGNAL_CORE == reason)
				    msg += "; exited due to a signal";
				return msg;
			}

```
This would allow better code reuse.

swebb2066 (commented):
The are a number of usages of apr_proc_wait in the test cases as well. It would be appropriate to use SubProcessFailure::makeMessage it you are willing.

sahvx655-wq:
I have updated the implementation to use SubProcessFailure for subprocess exit handling and pushed the changes for review.

PR #722, opened 2026-07-18
Fix size_t underflow in MDCPatternConverter::format() truncation

94xhn:
## Problem

`MDCPatternConverter::format()` computes the remaining space budget for the "full MDC" JSON output as:

```cpp
size_t remainingLength = info.getMaxLength() - 1;
```

`getMaxLength()` returns an `int`. When a pattern specifies a maximum field width of zero (e.g. `%.0J`, or `%mdc` configured with precision `0`), this evaluates as `0 - 1` on an unsigned `size_t`, wrapping around to `SIZE_MAX`. The subsequent guard `if (remainingLength < itemLength) break;` then never fires, so the truncation is silently defeated and the **entire** MDC map is written out instead of nothing — the opposite of what the pattern requested.

This is reachable through any normal `PatternLayout` conversion pattern using `%J`/`%.NJ` (registered via `RULES_PUT("J", MDCPatternConverter)` in `patternlayout.cpp`), as well as through `%mdc`. `patternparser.cpp`'s `DOT_STATE` stores the digits after `.` directly as `maxLength` with no lower-bound check, so `maxLength == 0` is a legitimate, reachable value, not a theoretical one.

The generic truncation path used elsewhere (`FormattingInfo::adjustField()`) already handles `maxLength == 0` safely (plain cast, no subtraction), so this "full MDC" branch's own hand-rolled counter is the actual outlier/bug, not an intentional deviation.

## Fix

Clamp before subtracting, matching the safe-cast pattern already used in `FormattingInfo::adjustField()`:

```cpp
size_t remainingLength = info.getMaxLength() > 0
    ? static_cast<size_t>(info.getMaxLength()) - 1
    : 0;
```

## Testing

Added `MDCTestCase::test3`, which exercises `%.0J` against an MDC containing one entry and asserts the output now correctly contains no MDC content, instead of the full (unbounded) map. Existing `test2` (`%.30J`) and default/`INT_MAX` cases are unaffected — verified the arithmetic is unchanged for all `maxLength > 0` values, only the `maxLength == 0` case changes (from wrap-around to correctly-clamped `0`).

### DIFF
--- a/src/test/cpp/mdctestcase.cpp
+++ b/src/test/cpp/mdctestcase.cpp
@@ -33,6 +33,7 @@ LOGUNIT_CLASS(MDCTestCase)
 	LOGUNIT_TEST_SUITE(MDCTestCase);
 	LOGUNIT_TEST(test1);
 	LOGUNIT_TEST(test2);
+	LOGUNIT_TEST(test3);
 	LOGUNIT_TEST_SUITE_END();
 
 public:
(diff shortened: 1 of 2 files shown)


src/test/cpp/mdctestcase.cpp
```diff
@@ -74,6 +75,16 @@ LOGUNIT_CLASS(MDCTestCase)
...
 		LOGUNIT_ASSERT_EQUAL(LOG4CXX_STR("INFO  MDC.LayoutTest - {\"key1\":\"value1\"} Message"), output);
 	}
+
+	/// A maximum field length of zero must suppress all MDC content instead of underflowing.
+	void test3()
+	{
+		MDC item1("key1", "value1");
+		LogString output;
+		PatternLayout l{ LOG4CXX_STR("%-5p %c - %.0J %m") };
+		l.format(output, std::make_shared<spi::LoggingEvent>(LOG4CXX_STR("MDC.LayoutTest"), Level::getInfo(), LOG4CXX_STR("Message"), spi::LocationInfo::getLocationUnavailable()));
```
swebb2066: This test passes even without the mdcptternconverter.cpp change. To create a failing test, use the following:
```
		pattern::MDCPatternConverter converter;
		bool leftAlign{false};
		const int minLength{0}, maxLength{0};
		converter.setFormattingInfo(std::make_shared<pattern::FormattingInfo>(leftAlign, minLength, maxLength));
		MDC item1("key1", "value1");
		auto e = std::make_shared<spi::LoggingEvent>(LOG4CXX_STR("MDC.LayoutTest"), Level::getInfo(), LOG4CXX_STR("Message"), spi::LocationInfo::getLocationUnavailable());
		LogString output;
		converter.format(e, output);
		LOGUNIT_ASSERT_EQUAL(LOG4CXX_STR(""), output);
```

swebb2066 (commented):
You also need to include `<log4cxx/pattern/mdcpatternconverter.h>` in mdctestcase.cpp

PR #723, opened 2026-07-20
Prevent supurious 2GB test failure

swebb2066:


rm5248:
I just tried this out on my VM, it continues to be killed by the OOM killer when the build system is low on memory.  I did try appending the data in a loop instead(so it only appended a smaller percentage at a time), but that did not work at all either.  I would _expect_ this to work correctly,  but I think it's a Linux memory overcommit problem.  I tried to fix it by manually allocating a buffer of the correct size using malloc, but it still was killed.

swebb2066:
Can you try running `sudo sysctl vm.overcommit_memory=2` before running the test?

rm5248:
Using the same setup(applications using a fair amount of RAM), setting the `vm.overcommit_memory` almost immediately causes the desktop to crash.  I also changed the `sysctl vm.overcommit_ratio=90`, that at least allowed the desktop to not crash immediately.  However, one tab in firefox kept crashing at that point and 0/68 tests in log4cxx passed(every single subprocess failed with an abort).  Changing the `overcommit_memory` back to 0 allowed the tests to run.

This can clearly be a very flaky test given the large amount of memory that needs to be allocated(especially since it needs to be contiguous), so if we push out a release with this set as default I would expect us to receive a number of bug reports.

swebb2066:
What do you suggest we do?
1. Remove the test
2. Make it conditional on a CMake parameter.
3. Some other test (I do not know of any) to determine if the out-of-memory killer will apply.

rm5248:
Option 2 seems like the best solution to me.  We can still leave it enabled on CI(since it seems to be stable there).

rm5248:
I just checked the latest commit, it looks good to me.  I can run it with low RAM and everything succeeds, and then turning on the full check will pass if I have enough RAM to run the test.

PR #730, opened 2026-08-17
Encode an invalid codepoint using a charset-specific character

swebb2066:
This PR centralises decoding logic and uses the Unicode replacement character codepoint to indicate a failure.

The other changes are:
* UTF-8 encoding is available (using CharsetEncoder::getUTF8Encoder()) when Log4cxx is built using LOG4CXX_CHARSET=utf-8.
(Prior to this PR,  when Log4cxx was built using LOG4CXX_CHARSET=utf-8, the CharsetEncoder::getUTF8Encoder() function returned a TrivialCharsetEncoder.)

* The Telnet appender uses the Unicode replacement character codepoint to indicate a message encoding failure. (Prior to this PR,  it used a '?' character).
* The ```encode3_1``` test case in ```transcodertestcase``` now works when LOG4CXX_CHAR=wchar_t (0xff is a valid UTF-16 codepoint). (Prior to this PR, the 0xff character was converted to a '?' by the ```LOGUNIT_ASSERT_EQUAL``` implementation for wchar_t).

### DIFF
--- a/.github/workflows/abi-compatibility.yml
+++ b/.github/workflows/abi-compatibility.yml
@@ -27,7 +27,6 @@ jobs:
   job:
     name: abi-check
     runs-on: ubuntu-latest
-    timeout-minutes: 38
     strategy:
       fail-fast: false
 
@@ -39,7 +38,6 @@ jobs:
 
     - name: 'Configure Dependencies - Ubuntu'
       run: |
-        sudo apt-get update
         sudo apt-get install -y g++-11 libapr1-dev libaprutil1-dev elfutils vtable-dumper universal-ctags
 
     - name: 'run CMake'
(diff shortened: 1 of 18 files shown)


rm5248:
>     * The Telnet appender uses the Unicode replacement character codepoint to indicate a message encoding failure. (Prior to this PR,  it used a '?' character).

Originally Telnet was ASCII only, although from reading RFC 5198 I think UTF-8 should be valid in most cases.  Should we have this a configurable option for the telnet appender?

swebb2066:
> Should we have this a configurable option for the telnet appender?

It already allows the encoder to be configured (e.g. to "US-ASCII"), but it calls CharsetEncoder::getUTF8Encoder() as the default.

PR #738, opened 2026-08-24
Log a 'Unknown option ...' warning for an obsolete/incorrect configuration file item

swebb2066:
This PR provides diagnostic help to a user with a configuration file problem.


src/main/include/log4cxx/spi/optionhandler.h
```diff
@@ -78,8 +78,7 @@ class LOG4CXX_EXPORT OptionHandler : public virtual helpers::Object
 		other may be activated only when #activateOptions is
 		called.
 		*/
-		virtual void setOption(const LogString& option,
-			const LogString& value) = 0;
+		virtual void setOption(const LogString& option, const LogString& value);
```
rm5248, line 81: I think removing the `= 0` to declare the function pure virtual is causing the ABI check to fail.

It's not entirely clear to me if this is OK or not, but I don't think it is.  According to the [KDE compatibility guidelines](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B):

> You can...
> Reimplement virtual functions defined in the primary base class hierarchy (that is, virtuals defined in the first non-virtual base class, or in that class's first non-virtual base class, and so forth) **if** it is safe that programs linked with the prior version of the library call the implementation in the base class rather than the derived one. This is tricky and might be dangerous. Think twice before doing it. Alternatively see below for a workaround.

PR #758, opened 2026-09-10
Remove unused PREFER_BOOST cmake option

swebb2066:
This PR removes the generation of the unused `boost-std-configuration.h` and search for boost during a build


rm5248:
This should probably be done as part of the ABI bump, in the unlikely event that somebody is using boost.  I think we're about ready to do the ABI bump anyway, so this is a good thing to do.

PR #761, opened 2026-09-12
Prevent a missed wake-up of the background thread

swebb2066:
This PR fixes the flaky ThreadUtility test by addressing the underlying missed notify.


swebb2066:
The test may remain flaky on MacOS-14 and MacOS-15 [due to this bug](https://github.com/llvm/llvm-project/issues/76807)

PR #764, opened 2026-09-20
Allow plugins to not be in log4cxx namespace

rm5248:
Plugins are currently required to be in the log4cxx namespace.  The code that loads the classes doesn't care what namespace the plugin is in(unlike how java packages work).

Modified the example to have the custom logger not in the log4cxx namespace.


swebb2066:
I believe I had already modified the example so it was no longer in the Log4cxx namespace. Are you sure this is necessary?

rm5248:
perhaps it would be more accurate to say that the macros must be in either the `log4cxx` namespace or you must have `using namespace log4cxx` in your code.  If you comment out the `using namespace LOG4CXX_NS;` in the current master it won't compile because of the namespace errors.

